PR #7, opened 2024-02-19
ID-27 Create-skeleton-API-application

fazemiah-moj:
# Purpose

Create skeleton API application based of Laminas Skeleton

## Approach

- Removed views/css/jss as they are not needed
- Add JsonModel dependency in controllers instead
- Dockerfiles use Alpine and multi-stage builds based of Sirius front-end
- files cleanup


service-api/Dockerfile
```diff
@@ -1,7 +1,58 @@
 FROM php:8.2-fpm-alpine AS app
 
-RUN mkdir /app
-WORKDIR /app
-COPY index.php .
+RUN apk --no-cache add postgresql-dev fcgi icu-dev autoconf linux-headers $PHPIZE_DEPS \
+  && docker-php-ext-configure intl \
+  && docker-php-ext-install pdo pdo_pgsql opcache intl sockets zip \
```
mattmachell, line 5: You probably won't need all of these, since intent is to use dynamo for storage and we won't be doing any zip work in this app. I'd remove all but opcache until you need them.

service-api/composer.json
```diff
@@ -0,0 +1,81 @@
+{
+    "name": "ministryofjustice/opg-paper-identity-api",
+    "description": "OPG Paper Identity API",
+    "authors": [
+        {
+            "name": "Office of the Public Guardian"
+        }
+    ],
+    "type": "project",
+    "license": "BSD-3-Clause",
```
mattmachell, line 10: Please make sure all MOJ work is MIT license.

service-api/config/autoload/local.php.dist
```diff
@@ -0,0 +1,16 @@
+<?php
+
+/**
+ * Local Configuration Override
+ *
+ * This configuration override file is for overriding environment-specific and
+ * security-sensitive configuration information. Copy this file without the
+ * .dist extension at the end and populate values as needed.
+ *
+ * NOTE: This file is ignored from Git by default with the .gitignore included
```
mattmachell, line 10: Given we use containers and configure by environment you can get rid of this file (and it's non-dist version). The idea being your config differences for development should be made at the docker compose level, not in php code files. We tend to follow https://12factor.net/ where we can.

service-api/public/img/favicon.ico
gregtyler, line 1: Can delete this favicon

service-api/public/img/laminas-logo.svg
gregtyler, line 1: Can delete this logo

service-api/data/cache/.gitkeep
gregtyler, line 1: Don't need this, since we store our cache in `/tmp/cache`

gregtyler (commented):
Looks good, just spotted a few files that we don't need.

service-api/data/cache/.gitkeep
fazemiah-moj: Thanks pushing up the updates now

gregtyler (approved):

PR #10, opened 2024-02-26
ID-30-Add-PHPUnit-tests-to-the-CI-pipeline

fazemiah-moj:
# Purpose

Run phpunit tests run in the CI pipeline in GitHub Actions via docker

## Learning

Adding APIproblem tool to fix broken tests due to no views/rendering change


.github/workflows/build.yaml
```diff
@@ -0,0 +1,21 @@
...
+
+jobs:
+  test-unit:
+    name: Unit Testing
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run Unit Tests Docker
+        run: make api-unit-test
```
gregtyler, line 21: This should probably be run inside `main.yml` rather than separately, so that if tests fail it will block the rest of the pipeline

service-api/module/Application/src/Module.php
gregtyler, line 1: I don't think these changes will be needed with the changes already in [config/autoload/global.php](https://github.com/ministryofjustice/opg-paper-identity/pull/10/files#diff-63c522171cdb1417a01afd58248c53f5d78765ebc1e3e7736c17c9366f4e7e08)

.github/workflows/build.yaml
fazemiah-moj: Ah right sure :)

ramchale (approved):

mattmachell (approved):

PR #23, opened 2024-03-05
Feature/id 34 donor id check

spinhalf-com:
# Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-front/module/Application/src/Contracts/OpgApiServiceInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Contracts;
+
+interface OpgApiServiceInterface
```
gregtyler, line 7: FWIW we don't tend to use this pattern when interfaces only have one concrete. Which isn't to say "don't use it" but more "don't feel obliged to use it".

service-front/module/Application/view/layout/phase_banner.phtml
gregtyler, line 1: I don't think this partial is being used?

service-front/module/Application/view/application/pages/donor_id_check.phtml
```diff
@@ -0,0 +1,100 @@
...
+            <li>
+                do they have the ID document with them?
+            </li>
+            <li>
+                Check the ID matches what is in Sirius (update as appropriate)
+            </li>
+        </ul>
+
+        <article class="moj-ticket-panel" aria-label="Sub navigation 1">
+            <section class="moj-ticket-panel__content moj-ticket-panel__content--blue" aria-label="Section 1">
```
gregtyler, line 32: I'm not sure why the example in the design system uses article/section and aria-labels, but they don't semantically makes sense here so we should just use divs.
```suggestion
        <div class="moj-ticket-panel">
            <div class="moj-ticket-panel__content moj-ticket-panel__content--blue">
```

service-front/Dockerfile
```diff
@@ -69,4 +69,4 @@ COPY phpcs.xml phpcs.xml
 COPY phpunit.xml phpunit.xml
 COPY psalm.xml psalm.xml
 
-FROM app AS production
+FROM app AS production
```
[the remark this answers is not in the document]
spinhalf-com: Thanks - I enabled it in PHPStorm now

service-front/module/Application/view/application/pages/donor_id_check_with_table.phtml
gregtyler, line 1: Sorry, just spotted: I don't think this file or `donor_lpa_check.phtml` are used?

service-front/module/Application/src/Factories/OpgApiServiceFactory.php
```diff
@@ -0,0 +1,25 @@
...
+class OpgApiServiceFactory implements FactoryInterface
+{
+    /**
+     * @param ContainerInterface $container
+     * @param string                          $requestedName
+     * @param array<mixed>|null               $options
+     */
+    public function __invoke(ContainerInterface $container, $requestedName, array $options = null): OpgApiService
+    {
+        $guzzleClient = new Client(['base-url' => 'http://api-web']);
```
gregtyler, line 21: Would still be good to move this to an env var but admittedly we can do that later

gregtyler (approved):

gregtyler (approved):

PR #53, opened 2024-03-22
feature/ID-61-Mock-NI-Number-verification

fazemiah-moj:
# Purpose

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-api/module/Application/src/Services/MockNinoService.php
+++ b/service-api/module/Application/src/Services/MockNinoService.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Services;
+
+use Application\Services\Contract\NINOServiceInterface;
+
+class MockNinoService implements NINOServiceInterface
+{
+
+    public function validateNINO(string $nino): string
+    {
+        if (!is_string($nino)) {
+            $serviceManager =
+            $this->logger->error(
+                'Invalid data passed to validateNino()',
+                [
+
+                ]
+            );
+
+        }
+        if (str_ends_with($nino, 'C')) {
+            return 'No Match';
+        }
+        elseif (str_ends_with($nino, 'D')) {
+            return 'Not Enough Details To Continue';
+        } else {
+            return 'Pass';
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


service-api/module/Application/src/Services/MockNinoService.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Application\Services;
+
+use Application\Services\Contract\NINOServiceInterface;
+
+class MockNinoService implements NINOServiceInterface
+{
+    public function validateNINO(string $nino): string
+    {
+        if (str_ends_with($nino, 'C')) {
+            return 'No Match';
```
gregtyler, line 14: These should return constant-defined codes (e.g. `NO_MATCH`) rather than text so that we're not tying content to code.

service-api/module/Application/src/Services/NinoAPIService.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Services;
```
gregtyler, line 5: Been thinking a bit more about this: I think all of these services should be in a separate namespace so that we have clear isolation. When we replace them with real integrations, things like authentication are going to be very different and will require support classes. Isolating them allows each to grow naturally.

So I'd suggest we end up with something like:
- `Application\Nino\Validator`
- `Application\Nino\ValidatorInterface`
- `Application\Nino\ValidatorFactory`

And then the mock should maybe live at:
- `Application\Mock\Validator`

Interested to hear what you think about that.

service-api/module/Application/src/Services/MockNinoService.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Application\Services;
+
+use Application\Services\Contract\NINOServiceInterface;
+
+class MockNinoService implements NINOServiceInterface
+{
+    public const NO_MATCH = 'Unable to verify NINO details';
+    public const NOT_ENOUGH_DETAILS = 'Not enough details to continue with this form of identification';
+    public const PASS = 'NINO check complete';
```
gregtyler, line 13: Just realised these should live on the interface since they're an essential part of the contract.

service-api/module/Application/src/Services/NinoAPIService.php
fazemiah-moj: Think it makes sense since we already have at 3/4 different types of these ID services itself

spinhalf-com (approved):

spinhalf-com (approved):

PR #68, opened 2024-04-10
feature/ID-90-Populate-API-endpoint-to-create-a-new-case

fazemiah-moj:
# Purpose

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-api/module/Application/src/Controller/IdentityController.php
+++ b/service-api/module/Application/src/Controller/IdentityController.php
@@ -4,11 +4,14 @@ declare(strict_types=1);
 
 namespace Application\Controller;
 
+use Application\Logging\Context;
+use Application\Mvc\Controller\Plugin\ValidatePayload\BadPayloadException;
 use Application\Nino\ValidatorInterface;
 use Application\DrivingLicense\ValidatorInterface as LicenseValidatorInterface;
 use Application\Passport\ValidatorInterface as PassportValidator;
 use Application\Fixtures\DataImportHandler;
 use Application\Fixtures\DataQueryHandler;
+use Application\Model\Entity\CaseData;
 use Laminas\Http\Response;
 use Laminas\Mvc\Controller\AbstractActionController;
 use Laminas\View\Model\JsonModel;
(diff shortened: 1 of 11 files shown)


service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -35,8 +37,29 @@ public function indexAction(): JsonModel
...
+                'dob'           => ['S' => $data["dob"]],
+                'verifyMethod'  => ['S' => $data["verifyMethod"]],
+                'lpas'          => ['SS' => $data['lpas']]
+            ];
+
+            $this->dataImportHandler->insertData('cases', $item);
+
+            return new JsonModel(['uuid' => $uuid]);
+        }
+        return new JsonModel(['error' => 'Invalid data']);
```
gregtyler, line 62: Should set the response status to 400 to indicate that the requested failed and it's the client's fault

service-api/module/Application/src/Fixtures/DataImportHandler.php
```diff
@@ -60,4 +60,17 @@ public function importData(array $data): void
...
+    public function insertData(string $tablename, array $item): void
+    {
+        $params = [
+            'TableName' => $tablename,
+            'Item' => $item
+        ];
+        try {
+            $this->dynamoDbClient->putItem($params);
+        } catch (AwsException $e) {
+            echo "Error: " . $e->getMessage();
```
gregtyler, line 73: Should import `LoggerInterface` and use its `error` method so that it's picked up by our automated alarms

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * DTO for holding data required to make new case entry post
+ * @psalm-suppress MissingConstructor
+ * Needed here due to false positive from Laminas’s uninitialised properties
+ */
+class CaseData implements JsonSerializable
+{
+    #[Validator(InArray::class, options: ["haystack" => ['passport', 'drivinglicense', 'nino']])]
+    private string $verifyMethod;
```
gregtyler, line 24: I don't think we should record this: we won't know it until at the point the case is created, and the user might change or select multiple options.

We'll need to store any verified documents on the record, but that will come in a later request

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -0,0 +1,98 @@
...
+    private string $lpa1;
+
+    #[Annotation\Validator(LPAValidator::class)]
+    private ?string $lpa2;
+
+    #[Annotation\Validator(LPAValidator::class)]
+    private ?string $lpa3;
+
+
+    private ?string $lpa4;
```
gregtyler, line 48: You can use `Explode` to allow an arbitrary-length array of LPA IDs:
```suggestion
    /**
     * @var string[]
     */
    #[Annotation\Validator(Explode::class, options: ['validator' => ['name' => LPAValidator::class]])]
    private array $lpas;
```

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -0,0 +1,98 @@
...
+        $instance->lpa3 = $data['lpas'][2] ?? 'NA';
+        $instance->lpa4 = $data['lpas'][3] ?? 'NA';
+
+        return $instance;
+    }
+
+    /**
+     * @psalm-suppress PossiblyUnusedMethod
+     */
+    public function jsonSerialize(): mixed
```
gregtyler, line 76: Is this serialization used? If so, it should be covered by a test. If not, we can remove it.

service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -35,8 +37,29 @@ public function indexAction(): JsonModel
         return new JsonModel();
     }
 
-    public function createAction(): void
+    public function createAction(): JsonModel
```
gregtyler, line 40: This needs tests

service-api/module/Application/src/Fixtures/DataImportHandler.php
```diff
@@ -60,4 +60,17 @@ public function importData(array $data): void
             }
         }
     }
+
+    public function insertData(string $tablename, array $item): void
```
gregtyler, line 64: This needs tests

service-api/module/Application/src/Validators/LPAValidator.php
```diff
@@ -0,0 +1,36 @@
...
+        return $this->lpaValidity($value);
+    }
+
+    private function lpaValidity(?string $lpa): bool
+    {
+        if ($lpa === null || $lpa === '') {
+            return true;
+        }
+            /** @var string $lpa */
+        if (1 == preg_match('/M(-([0-9A-Z]){4}){3}/', $lpa)) {
```
gregtyler, line 30: Should use strict equality:
```suggestion
        if (false !== preg_match('/M(-([0-9A-Z]){4}){3}/', $lpa)) {
```

gregtyler (approved):

service-api/module/Application/src/Validators/LPAValidator.php
gregtyler: Sorry, your solution `1 === preg_match()` is much more accurate because mine would have evaluated true for non-matches too 😅

PR #75, opened 2024-04-22
feature/ID-82-Create-PHP-service-to-generate-KBV-questions-and-answers

fazemiah-moj:
# Purpose

This PR combines both ID-82, ID-78 & ID-79

To verify functionality:

1 - Create a new case via hitting this endpoint: http://localhost:8001/identity/create

example json payload expected:

`{
	"firstName": "Joe",
    "lastName": "Test",
    "personType": "Donor",
    "dob": "1993-01-09",
    "lpas" : ["M-XOXO-YAGA-XIXL", "M-YOYO-AJAX-3LLP"]
}`

2. Take note of UUID sent back, can also verify this record created here http://localhost:8001/identity/details?uuid=UUID

3. Hit http://localhost:8001/cases/UUID/kbv-questions - this will generate the questions if it does not exist, again can verify this by hitting /details/ endpoint above
 
4. Refresh step 3 to check it is the same questions that is returned back

5. KBV verifications can be tested with any specific case, post generating the questions here: http://localhost:8001/cases/ebdaacbb-f31b-416b-82bc-7ef2ccf709dd/kbv-answers

example json payload expected:

```
{
    "answers": {
        "one": "£5.99",
        "two": "Liberty Trust Bank",
        "three": "July",
        "four": "SafeDrive Insurance"
    }
}
```

You will need to verify what the correct answers to the questions generated for the specific case by referencing https://github.com/ministryofjustice/opg-paper-identity/blob/849a5692ecb7da001cd341f35084334a743ddcfe/service-api/module/Application/src/Mock/KBV/KBVService.php#L52-L52
 

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/KBV/KBVServiceInterface.php
```diff
@@ -0,0 +1,19 @@
...
+interface KBVServiceInterface
+{
+    /**
+     * @return array{
+     *  question: string,
+     *  prompts: string[],
+     *  answer: string
+     * }[]
+     */
+    public function getKBVQuestions(): array;
```
gregtyler, line 16: This method isn't used externally so can probably be removed from the interface and made `private` if kept in implementations.

Appreciate this method was part of the interface that was on the ticket, but if the return value isn't useful then there's no point keeping it.

service-api/module/Application/src/Fixtures/DataImportHandler.php
```diff
@@ -77,4 +78,46 @@ public function insertData(string $tablename, array $item): void
...
+        try {
+            $this->updateItemAttributeByKey('cases', $idKey, $attrName, $attrType, $attrValue);
+        } catch (AwsException $e) {
+            $this->logger->error('Unable to update data [' . $e->getMessage() . '] for case' . $uuid, [
+                'data' => [$attrName => $attrValue]
+            ]);
+        }
+    }
+
+    public function updateItemAttributeByKey(
```
gregtyler, line 101: I don't think this function is used anywhere other than `updateCaseData`, could they just be merged together?

fazemiah-moj: Yeah makes sense to condense this, thought might have other variations of use but not really needed

gregtyler (approved):

PR #80, opened 2024-05-02
feature/ID-107-Create-a-mock-for-yoti-API-for-local-dev-use

fazemiah-moj:
# Purpose

Endpoints specified and payload/response structures set out and ready to be tested.

Some open questions/clarifications pending:

* Also docs do not make it clear where the **application_profile** sits within the session creation payload structure, which is an open question 
* Post validation - currently applying a generic pattern for matching UK ones. Is there a more specific Post Office version? how about international ones?
* Currently unclear what /idverify/v1/sessions/{sessionId}/instructions/pdf should actually return. Docs indicate this would be used to get and create the PDF letter, but how is this possible with no data sending back on a 200? Question added on our Word document

Fixes ###-####

## Approach

## Learning

## Checklist


docker/yoti/data/branches.json
gregtyler, line 1: I don't think this file is used?

docker-compose.yml
```diff
@@ -14,6 +14,20 @@ services:
     volumes:
       - ./docker/sirius:/opt/imposter/config
 
+  yoti-mock:
+    image: outofcoffee/imposter
+    restart: unless-stopped
+    ports:
+      - 8002:80
```
gregtyler, line 21: No need expose ports under normal operation
```suggestion
```

docker/yoti/data/tsconfig.json
gregtyler, line 1: We don't use TypeScript so shouldn't need a tsconfig file

docker/yoti/openapi.yml
```diff
@@ -0,0 +1,640 @@
...
+                    - $ref: "#/components/schemas/AddressFormat2"
+                    - $ref: "#/components/schemas/AddressFormat3"
+                    - $ref: "#/components/schemas/AddressFormat4"
+
+    ID_DOCUMENT:
+      type: object
+      properties:
+        type:
+          type: string
+          enum: [ ID_DOCUMENT, SUPPLEMENTARY_DOCUMENT ]
```
gregtyler, line 338: ID_DOCUMENTs only have one option for type
```suggestion
          enum: [ ID_DOCUMENT ]
```

docker/yoti/openapi.yml
```diff
@@ -0,0 +1,640 @@
...
+                      type: string
+                  document_types:
+                    type: array
+                    items:
+                      type: string
+                      enum: [PASSPORT, DRIVING_LICENCE, NATIONAL_ID, RESIDENCE_PERMIT]
+
+    SUPPLEMENTARY_DOCUMENT:
+      type: object
+      properties:
```
gregtyler, line 363: Should add `type` here too:
```suggestion
      properties:
        type:
          type: string
          enum: [ SUPPLEMENTARY_DOCUMENT ]
```

gregtyler (commented):
A couple of minor comments but otherwise this looks really good based on the information we have

docker/yoti/openapi.yml
fazemiah-moj: Ah missed that one, good spot :)

gregtyler (approved):

PR #82, opened 2024-05-08
feature/ID-135-Create-YotiService-Class-Endpoint-Stubs

fazemiah-moj:
# Purpose

Mostly an outline of Yoti Service class functionality and endpoints so we can complete with respective upcoming tickets. This also needs Yoto mock to be merged in.

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/docker-compose.env
gregtyler, line 1: You'll find you've got a conflict with main here because we switched these to all be in alphabetical order.

service-api/docker-compose.env
```diff
@@ -1,5 +1,7 @@
 MOCK_NINO_API=true
 MOCK_DRIVINGLICENSE_API=true
+MOCK_YOTI=true
+LOCAL_YOTI_BASE_URL=
```
gregtyler, line 4: I don't think this needs a `LOCAL_` prefix? Other similar ones don't (e.g. `DRIVINGLICENSE_BASE_URL`)
```suggestion
YOTI_BASE_URL=
```

service-api/module/Application/src/Mock/Yoti/YotiService.php
gregtyler, line 1: In this instance I don't think we need a PHP-mock for Yoti since we can run the remote mock you've created with OpenAPI locally (we need it for other providers because we can't run their sandbox environments locally)

service-api/module/Application/config/module.config.php
```diff
@@ -196,6 +198,46 @@
                     ],
                 ],
             ],
+            'find_postoffice_branches' => [
+                'type'    => Segment::class,
+                'options' => [
+                    'route'    => '/yoti/post-office-branches',
```
gregtyler, line 204: I think it would be good to remove "yoti" from the URLs here, since it ties us a bit to the supplier. I think it's fine in file/class names but since the URLs form a contract they should describe what they do rather than how they do it.

mattmachell: Yeah, if you want to keep it abstract` /counter-service/branches` is more generalised.

service-api/module/Application/src/Mock/Yoti/YotiService.php
fazemiah-moj: Ah so would this be that the default for local or dev environments would be whatever port we have the openAPI mock unless, unless it's specifically production?

gregtyler: Yep, so:
- Local envs would use `Application\Yoti\YotiService` pointing to `yoti-mock:8002`
- Non-prod envs would use `Application\Yoti\YotiService` pointing to a hosted-in-AWS version of the Yoti mock
- Prod env would use `Application\Yoti\YotiService` pointing to the real prod server

Whereas with some of our other integrations:
- Local envs would use a mock service
- Non-prod envs would either use a mock service or a real service pointing to the sandbox version of the integration
- Prod env would use a real service pointing to the real prod server

service-api/docker-compose.env
```diff
@@ -9,5 +9,8 @@ MOCK_DRIVINGLICENSE_API=true
 MOCK_KBV_API=true
 MOCK_NINO_API=true
 MOCK_PASSPORT_API=true
+MOCK_YOTI=true
```
gregtyler, line 12: I don't think this env var does anything?
```suggestion
```

service-api/phpunit.xml
```diff
@@ -37,5 +37,7 @@
         <env name="AWS_DYNAMODB_TABLE_NAME" value="cases" />
         <env name="MOCK_KBV_API" value="true" />
         <env name="EXPERIAN_BASE_URL" value="EXPERIAN_BASE_URL" />
+        <env name="YOTI_MOCK" value="true" />
```
gregtyler, line 40: This doesn't do anything
```suggestion
```

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    public function __construct(
+        private readonly YotiServiceInterface $yotiService
+    ) {
+    }
+
+    public function findPostOfficeAction(string $postCode): JsonModel
+    {
+        $branches = [];
+        $branches["locations"] = $this->yotiService->postOfficeBranch($postCode);
```
gregtyler, line 26: This isn't covered by tests, so hard to confirm, but I don't think this matches the structure expected in [ID-137](https://opgtransform.atlassian.net/browse/ID-137).

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -0,0 +1,59 @@
...
+        return new JsonModel($branches);
+    }
+    public function createSessionAction(array $sessionData): JsonModel
+    {
+        $data = [];
+        $data['response'] = $this->yotiService->createSession($sessionData);
+        return new JsonModel($data);
```
gregtyler, line 33: I don't think we should return the data directly from Yoti's API, because then we end up with the front-end having a dependency on an external API. For now maybe worth just returning empty/204 until we have a need for any data to be passed back to the frontend.
```suggestion
        return new JsonModel(null);
```

We'll also need to save the session ID from the response to `CaseData` so that we can retrieve it again in the future.

service-api/module/Application/config/module.config.php
```diff
@@ -206,6 +208,46 @@
...
+                    'defaults' => [
+                        'controller' => Controller\YotiController::class,
+                        'action'     => 'createSession',
+                    ],
+                ],
+            ],
+            'retrieve_yoti_status' => [
+                'type'    => Segment::class,
+                'options' => [
+                    'route'    => '/counter-service[/:sessionId]/retrieve-status',
```
gregtyler, line 234: This URL should accept a case UUID, retrieve the session ID from its CaseData, and use that in the Yoti API. Otherwise the frontend would need to be aware of the session ID, but the frontend is stateless.
```suggestion
                    'route'    => '/counter-service/:uuid/retrieve-status',
```

service-api/module/Application/config/module.config.php
```diff
@@ -206,6 +208,46 @@
...
+                    'defaults' => [
+                        'controller' => Controller\YotiController::class,
+                        'action'     => 'getSessionStatus',
+                    ],
+                ],
+            ],
+            'retrieve_pdf_letter' => [
+                'type'    => Segment::class,
+                'options' => [
+                    'route'    => '/counter-service[/:sessionId]/retrieve-letter',
```
gregtyler, line 244: Same here:
```suggestion
                    'route'    => '/counter-service/:uuid/retrieve-letter',
```

service-api/module/Application/src/Controller/YotiController.php
fazemiah-moj: Created this purely as a stub/placeholder before ID-137 existed, plan is to tackle this next

service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -54,7 +54,10 @@ public function createAction(): JsonModel
         if ($validator->isValid()) {
             $caseData->id = strval(Uuid::uuid4());
 
+            //todo do get insertData() to return a true or false and if false
```
gregtyler, line 57: Is this still supposed to be here? John's looking at error capture on #114 
```suggestion
```

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -0,0 +1,83 @@
...
+class YotiService implements YotiServiceInterface
+{
+    public function __construct(
+        public readonly Client $client
+    ) {
+    }
+
+    /**
+     * @param string $postCode
+     * @return array
```
gregtyler, line 25: It would be good to expand these returns to indicate the shape of the array coming back (e.g. `@return array{fad_code: string, name: string, address: string[], postcode: string}[]`)

Same for other responses

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -0,0 +1,83 @@
...
+        try {
+            $results = $this->client->post('/idverify/v1/lookup/uk-post-office', [
+                'json' => ['SearchString' => $postCode],
+            ]);
+            if ($results->getStatusCode() !== Response::STATUS_CODE_200) {
+                throw new YotiException($results->getReasonPhrase());
+            }
+            return json_decode(strval($results->getBody()), true);
+        } catch (GuzzleException $e) {
+            return ["error" => $e->getMessage()];
```
gregtyler, line 41: I don't like returning errors in the same shape as successes because it makes it hard to distinguish the two. Maybe it would make more sense to throw e.g. `new YotiException("A connection error occurred", previous: $e)`?

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -0,0 +1,91 @@
...
+    public function findPostOfficeAction(): JsonModel
+    {
+        $branches = [];
+        $data = json_decode($this->getRequest()->getContent(), true);
+
+        if (! isset($data["search_string"])) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(['error' => 'Missing postCode']);
+        }
+        $branchList = $this->yotiService->postOfficeBranch($data["search_string"]);
```
gregtyler, line 38: `postOfficeBranch` throws exceptions but they aren't caught here at all, which means they'll just show up as generic 500s. We should at least log the error message so we can debug issues.

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -0,0 +1,91 @@
...
+     * @psalm-suppress PossiblyInvalidArgument
+     */
+    public function findPostOfficeAction(): JsonModel
+    {
+        $branches = [];
+        $data = json_decode($this->getRequest()->getContent(), true);
+
+        if (! isset($data["search_string"])) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(['error' => 'Missing postCode']);
```
gregtyler, line 36: Should use `Problem`"
```suggestion
            return new JsonModel(new Problem('Missing postCode'));
```

service-api/module/Application/src/Yoti/YotiService.php
fazemiah-moj: This would be the case in the controller when we do the formatting for our needs. The YotiService in this case would return the same format specified by Yoti which would be like

 `* @return array{
 *     branches: array{
 *         type: string,
 *         fad_code: string,
 *         name: string,
 *         address: string,
 *         postcode: string,
 *         location: array{
 *             latitude: float,
 *             longitude: float
 *         }
 *     }[]
 * }
 * `

I suppose?

gregtyler: Yeah that looks good. I think it's worth including that in the annotation so we have code-level documentation and type-hinting.

docker/yoti/data/branches.json
```diff
@@ -0,0 +1,37 @@
...
+    {
+      "type": "UK_POST_OFFICE",
+      "fad_code": "12345676",
+      "name": "Hampstead",
+      "address": "66 High Street, Hampstead Heath, LONON",
+      "postcode": "NW3 6LR",
+      "location": {
+        "latitude": 52.22864,
+        "longitude": -0.26762
+      }
```
ramchale, line 34: I'm not sure if it's going to confuse testing further down the line that this has the same location at the first branch?

ramchale (approved):

ramchale (approved):

PR #92, opened 2024-05-17
Feature/id 141 refactor form processor

spinhalf-com:
# Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -69,4 +69,16 @@ public function isValid(): bool
             ->setData(get_object_vars($this))
             ->isValid();
     }
+
+    public function toArray(): array
```
gregtyler, line 73: At a minimum we should define the array shape for Psalm. Explicit getters (`getPersonType`) would be better still, but array shape will at least catch errors.
```suggestion
    /**
     * @returns array{personType: "donor"|"certificateProvider", firstName: string, lastName: string, dob: string, address: string[], lpas: string[]}
     */
    public function toArray(): array
```

service-front/module/Application/src/Controller/CPFlowController.php
```diff
@@ -5,25 +5,19 @@
...
-use Laminas\Form\Annotation\AttributeBuilder;
-use Application\Forms\NationalInsuranceNumber;
 
 class CPFlowController extends AbstractActionController
 {
     protected $plugins;
     public function __construct(
         private readonly OpgApiServiceInterface $opgApiService,
-        private readonly FormProcessorService $formProcessorService,
+        private readonly FormProcessorHelper $formProcessorHellper,
```
gregtyler, line 20: Nit: spelling mistake on this property name
```suggestion
        private readonly FormProcessorHelper $formProcessorHelper,
```

service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -157,14 +155,18 @@ public function nationalInsuranceNumberAction(): ViewModel
...
+                $uuid,
                 $this->getRequest()->getPost(),
                 $form,
-                $view,
                 $templates
             );
-        }
+            foreach ($formProcessorResponseDto->getVariables() as $key => $variable) {
+                $view->setVariable($key, $variable);
+            }
```
gregtyler, line 166: I think you could use `setVariables` here, which would simplify the controller code quite a lot
```suggestion
            $view->setVariables($formProcessorResponseDto->getVariables());
```

service-front/module/Application/src/Helpers/DTO/FormProcessorRequestDto.php
gregtyler, line 1: Is this Request DTO used anywhere?

service-front/module/Application/src/Helpers/DTO/FormProcessorResponseDto.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function getForm(): FormInterface
+    {
+        return $this->form;
+    }
+
+    public function getResponseData(): array
+    {
+        return $this->responseData;
+    }
```
gregtyler, line 33: I can't work out what this is for, and can't see it being used anywhere. Perhaps removable now?

service-front/module/Application/view/application/pages/post_office/find_post_office.twig
```diff
@@ -1,33 +1,42 @@
...
+                        <span class="govuk-visually-hidden">Error:</span>
+                        {{ form.get('selected_postcode').messages | join(', ', ' and ') }}
+                    </p>
+                {% endif %}
+                <fieldset class="govuk-fieldset">
                     <div class="govuk-radios" data-module="govuk-radios">
+                        <div class="govuk-radios__item">
+                            <input
+                                    class="govuk-radios__input"
+                                    id="default"
```
gregtyler, line 30: Should probably be ID `postcode`, otherwise the label won't link up to it properly
```suggestion
                                    id="postcode"
```

service-front/web/main.scss
```diff
@@ -24,3 +24,26 @@ $govuk-assets-path: "../assets/";
...
+//{#/* identical to box height, or 132% */#}
+//
+//{#/* Tag/Red Text */#}
+//{#color: #2A0B06;#}
+//
+//
+//{#/* Inside auto layout */#}
+//{#flex: none;#}
+//{#order: 0;#}
+//{#flex-grow: 0;#}
```
gregtyler, line 49: Presumably all deletable?
```suggestion
```

gregtyler (commented):
I see what you mean about it potentially being too abstracted. The methods in `FormProcessorHelper` don't really have a clear reason to exist, since they're not being reused and they don't have particularly similar logic.

I guess a lot of the code in there could be moved into the controller it's used by, and then potentially simplified with some helper methods/traits for common bits of functionality. Would probably be worth in that case dividing up the controllers a bit further.

But that's all much of a muchness: what we have now is at least a pattern, and it works

service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -186,14 +186,19 @@ public function drivingLicenceNumberAction(): ViewModel
...
                 $this->getRequest()->getPost(),
                 $form,
-                $view,
                 $templates
             );
-        }
 
+            foreach ($formProcessorResponseDto->getVariables() as $key => $variable) {
+                $view->setVariable($key, $variable);
+            }
```
gregtyler, line 198: Looks like there are a few more loops that could benefit from the `setVariables($dto->getVariables())` treatment

service-front/module/Application/src/Helpers/DTO/FormProcessorResponseDto.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Application\Helpers\DTO;
+
+use Laminas\Form\FormInterface;
+
+class FormProcessorResponseDto
+{
+    public function __construct(
+        private string $uuid,
+        private FormInterface $form,
+        private array $responseData,
```
gregtyler, line 14: I think you've removed the getter for this, but the property is still here and I don't think it's doing anything

gregtyler (approved):

PR #104, opened 2024-05-28
Return objects in API responses

gregtyler:
## Purpose

The general goal here is that, rather than having to define the response shape in every controller action:
```php
$case = new CaseData(...);

return JsonModel([
  'id' => $case->getId(),
  // etc.
]);
```

You can just return the object:
```php
$case = new CaseData(...);

return JsonModel($case);
```

This has a few benefits:
- It ensures the data shape is consistent where multiple endpoints return the same shape (in particular, errors)
- It allows typed objects to be passed through methods more easily, rather than arrays that may or may not contain expected data (e.g. how DataQueryHandler->getCaseByUUID now returns CaseData)
- In turn, that forces us to define our data models better (I found that `kbvQuestions` and `documentComplete` were missing from `CaseData`)
- It will let us annotate the OpenAPI spec directly from the objects with [swagger-php](https://zircote.github.io/swagger-php/)

#minor

## Approach

I made the objects we want to return JsonSerializable and introduced a new JsonModel variant that uses that method. That seemed easier than supporting the official JsonModel class, which only handles Traversables, because it's much more verbose to make a class Traversable.

## Checklist

  * N/A
  * N/A
  * N/A
* [x] The team have tested these changes
  * N/A

### DIFF
--- a/service-api/module/Application/src/Controller/IdentityController.php
+++ b/service-api/module/Application/src/Controller/IdentityController.php
@@ -11,10 +11,11 @@ use Application\KBV\KBVServiceInterface;
 use Application\Fixtures\DataImportHandler;
 use Application\Fixtures\DataQueryHandler;
 use Application\Model\Entity\CaseData;
+use Application\Model\Entity\Problem;
+use Application\View\JsonModel;
 use Laminas\Form\Annotation\AttributeBuilder;
 use Laminas\Http\Response;
 use Laminas\Mvc\Controller\AbstractActionController;
-use Laminas\View\Model\JsonModel;
 use Ramsey\Uuid\Uuid;
 
 /**
@@ -69,11 +70,10 @@ class IdentityController extends AbstractActionController
 
         $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
 
-        return new JsonModel([
-            'type' => 'HTTP400',
-            'title'  => 'Invalid data',
-            'error' => $validator->getMessages(),
-        ]);
+        return new JsonModel(new Problem(
+            'Invalid data',
+            extra: ['errors' => $validator->getMessages()],
+        ));
     }
 
     public function detailsAction(): JsonModel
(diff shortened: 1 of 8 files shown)


service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -69,11 +70,10 @@ public function createAction(): JsonModel
 
         $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
 
-        return new JsonModel([
-            'type' => 'HTTP400',
-            'title'  => 'Invalid data',
-            'error' => $validator->getMessages(),
-        ]);
+        return new JsonModel(new Problem(
+            'Invalid data',
```
fazemiah-moj, line 74: Do we not need to explicitly specify this as a title?

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -45,6 +47,13 @@ class CaseData
     #[Annotation\Validator(Explode::class, options: ['validator' => ['name' => LpaUidValidator::class]])]
     public array $lpas;
 
+    public ?string $kbvQuestions = null;
+
+    #[Annotation\Required(false)]
+    #[Annotation\Validator(IsType::class, options: ['type' => 'boolean'])]
+    #[Annotation\Validator(NotEmpty::class, options: [NotEmpty::NULL])]
```
fazemiah-moj, line 54: Maybe i'm missing something, this looks like it would have the same affect as line 52?

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -71,18 +80,32 @@ public static function fromArray(mixed $data): self
...
             'firstName' => $this->firstName,
             'lastName' => $this->lastName,
             'dob' => $this->dob,
             'address' => $this->address,
             'lpas' => $this->lpas,
+            'documentComplete' => $this->documentComplete,
         ];
+
+        if ($this->kbvQuestions !== null) {
+            $arr['kbvQuestions'] = $this->kbvQuestions;
```
fazemiah-moj, line 101: Think kvbQuestions was initially missing as we had not started on the questions part yet, also because I could not think of what validation we would do here, did not add this subsequently but it makes sense to be here for completeness if we go ahead to pass CaseData as a whole object to the views

fazemiah-moj (approved):
Looks good Greg, JsonModel out of the box seemed a little problematic, hadn't crossed my mind that we could customize it. 

This would definitely save us some time and reduce errors/inconsistencies in output, thanks 👍

service-api/module/Application/src/Model/Entity/CaseData.php
gregtyler: It's a quirk of laminas-form that I don't fully understand: even if you say the field isn't required, it still seems to apply the `NotEmpty` filter and fails in surprising ways.

service-api/module/Application/src/Controller/IdentityController.php
gregtyler: Since it's the first argument in the constructor it knows it's the title. `extra` only needs to be specified because we're skipping the other arguments

PR #107, opened 2024-05-30
ID-136-Make-date-of-birth-optional-in-create-case-endpoint-in-API

fazemiah-moj:
This PR makes dob optional, unfortunately, seems the Validator does not accept an empty value and thus empty value matching, thus I resorted to defaulting it to "NA"

Also improved error handling to notify when a db save might fail

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -63,7 +63,16 @@ public function createAction(): JsonModel
...
             ];
 
-            $this->dataImportHandler->insertData($item);
+            $insert = $this->dataImportHandler->insertData($item);
+
+            if (! $insert) {
+                $this->getResponse()->setStatusCode(Response::STATUS_CODE_422);
+                return new JsonModel(new Problem(
+                    'Data save failed / invalid data format',
+                    extra: ['errors' => 'Please see logs'],
```
gregtyler, line 72: We don't need to tell the front end to check the logs, this is implicit in our processes
```suggestion
```

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -66,7 +67,7 @@ public static function fromArray(mixed $data): self
         $instance->personType = $data['personType'];
         $instance->firstName = $data['firstName'];
         $instance->lastName = $data['lastName'];
-        $instance->dob = $data['dob'];
+        $instance->dob = $data['dob'] ?? 'NA';
```
gregtyler, line 70: If the field is nullable, this should fallback to null rather than a string
```suggestion
        $instance->dob = $data['dob'] ?? null;
```

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -24,10 +24,11 @@ class CaseData implements JsonSerializable
     #[Validator(NotEmpty::class)]
     public string $personType;
 
-    #[Validator(Regex::class, options: ["pattern" => "/^[0-9]{4}-[0-9]{2}-[0-9]{2}$/", "messages" => [
+    #[Annotation\Required(false)]
+    #[Validator(Regex::class, options: ["pattern" => "/^NA$|^[0-9]{4}-[0-9]{2}-[0-9]{2}$/", "messages" => [
```
gregtyler, line 28: An empty value should be null or missing, not the string NA.
```suggestion
    #[Validator(Regex::class, options: ["pattern" => "/^[0-9]{4}-[0-9]{2}-[0-9]{2}$/", "messages" => [
```

service-api/module/Application/src/Model/Entity/CaseData.php
fazemiah-moj: My initial attempt was to have this as null as above, however it seems while this satisfies validator, dynanomdb does not like a null value, but can't find a reason why it wouldn't. Will dig abit more into this

spinhalf-com (approved):

PR #121, opened 2024-06-13
ID-116-Implement-secrets-manager-client-with-caching

fazemiah-moj:
## Purpose

Implements Secret Manager client with apcu caching so secrets are fetched from cache first when available.

This also implements https://opgtransform.atlassian.net/browse/ID-171 which uses this client

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Aws/Secrets/AwsSecretsCache.php
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * @throws ExceptionInterface
+     * @throws InvalidSecretResponseException
+     */
+    public function getValue(string $name): string
+    {
+        $name = $this->prefix . '/' . $name;
+        $key = self::NAMESPACE_AWS . ':' . $name;
+
+        /*if ($this->storage->hasItem($key)) {
```
mattmachell, line 32: Worth just deleting this if you're not using it.

mattmachell: But feels like you should use it, because otherwise you always go to the network for secret values.

fazemiah-moj: Ah yes, I only commented this out for debugging when keys were not being read properly due to a formatting issue. Will update this now :)

service-api/module/Application/test/Aws/Secrets/AwsSecretTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $this->sut->getValue();
+    }
+
+    public function testGetValueReturnsValueFromCache(): void
+    {
+        $cache = $this->createMock(AwsSecretsCache::class);
+        $cache->expects(self::once())
+            ->method('getValue')
+            ->with('test')
+            ->willReturn('secretValue');
```
mattmachell, line 40: Is `AwsSecretsCache->getValue` the right level to mock this at? At the moment you can't test if it goes to network or hits the cache. Just that the cache returns a value.

service-api/module/Application/src/Yoti/Http/RequestSigner.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Yoti\Http;
+
+use Application\Aws\Secrets\AwsSecret;
+use Yoti\Http\Exception\RequestSignException;
+
+/**
+ * Suppress pslam error til class is used
+ * @psalm-suppress UnusedClass
+ */
+class RequestSigner
```
mattmachell, line 14: No tests around request signer?

service-api/module/Application/test/TestData/test-private-key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
mattmachell, line 1: Private keys can trigger various security tooling looking for leaked secrets (even if they are obviously test ones). Might be worth making a make target that generates these instead of committing, just saves time with queries about false positives later.

mattmachell (approved):

PR #127, opened 2024-06-27
Feature/id 139 remove lpa stub api method

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


scripts/localstack/init/localstack_init.sh
```diff
@@ -8,6 +8,14 @@ awslocal dynamodb create-table \
     AttributeName=id,KeyType=HASH \
   --provisioned-throughput ReadCapacityUnits=1000,WriteCapacityUnits=1000
 
+awslocal dynamodb update-table \
+    --table-name identity-verify \
+    --attribute-definitions AttributeName=lpas,AttributeType=SS \
+    --global-secondary-index-updates \
+        "[{\"Create\":{\"IndexName\": \"lpas-index\",\"KeySchema\":[{\"AttributeName\":\"lpas\",\"KeyType\":\"HASH\"}], \
+        \"ProvisionedThroughput\": {\"ReadCapacityUnits\": 20, \"WriteCapacityUnits\": 10 }, \
+        \"Projection\":{\"ProjectionType\":\"ALL\"}}}]"
```
gregtyler, line 17: Is this index needed? I couldn't see it being used.

If it is needed, you'll also need to add it to [the infrastructure code](https://github.com/ministryofjustice/opg-sirius-infrastructure/blob/8ace6e6a5998a281c62aef5cddf8c808816c337c/environment/modules/lpa_identity_dynamodb/dynamodb.tf).

service-front/module/Application/src/Helpers/LpaFormHelper.php
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+
+    public function getLpaTypeFromSiriusResponse(array $siriusCheck): string
+    {
+        try {
+            return $siriusCheck['opg.poas.lpastore']['lpaType'];
+        } catch (\Exception $exception) {
+            return $exception->getMessage();
+        }
```
gregtyler, line 134: The only error this is likely to throw is a PHP Warning if the key doesn't exist, and PHP Warnings can't be caught with a try..catch.

Instead, you probably want to do something like:
```suggestion
        return $siriusCheck['opg.poas.lpastore']['lpaType'] ?? '';
```
which will return an empty string if the key didn't exist.

service-front/module/Application/src/Helpers/LpaFormHelper.php
```diff
@@ -0,0 +1,263 @@
...
+        return new LpaFormHelperResponseDto(
+            $uuid,
+            $form,
+            [
+                'lpa_response' => $result
+            ],
+        );
+    }
+
+    public function getCpAddressFromSiriusResponse(array $siriusCheck): array
```
gregtyler, line 119: It looks like this and every other method below should be private, since `findLpa` is the only one used by anything external.
```suggestion
    private function getCpAddressFromSiriusResponse(array $siriusCheck): array
```

service-front/module/Application/src/Helpers/LpaFormHelper.php
```diff
@@ -0,0 +1,263 @@
...
+
+            $idCheck = $this->compareCpRecords($detailsData, $siriusCheck);
+            $statusCheck = $this->checkStatus($siriusCheck);
+            $channelCheck = $this->checkChannel($siriusCheck);
+
+            if ($statusCheck['error'] === true) {
+                $result['status'] = $statusCheck['status'];
+                $result['message'] = $statusCheck['message'];
+            } elseif ($idCheck['error'] === true) {
+                $result['status'] = 'no match';
```
gregtyler, line 65: I don't think it's a good idea to overwrite external status data, otherwise we can't reliably display the original status of the case.

service-front/module/Application/src/Helpers/LpaFormHelper.php
```diff
@@ -0,0 +1,263 @@
...
+                ];
+                $result['data'] = [
+                    "case_uuid" => $uuid,
+                    "LPA_Number" => $form->get('lpa')->getValue(),
+                    "Type_Of_LPA" => $this->getLpaTypeFromSiriusResponse($siriusCheck),
+                    "Donor" => $this->getDonorNameFromSiriusResponse($siriusCheck),
+                    "Status" => $statusCheck['status'],
+                    "CP_Name" => $detailsData['firstName'] . " " . $detailsData['lastName'],
+                    "CP_Address" => $detailsData['address']
+                ];
```
gregtyler, line 82: I'm not really clear what the difference is between "data" and "additional_data". Could `LpaFormHelperResponseDto` be expanded to have additional OOP structure for this information, rather than using an unstructured array?

service-front/module/Application/src/Services/SiriusApiService.php
```diff
@@ -99,11 +113,25 @@ public function checkAuth(RequestInterface $request): bool
      */
     public function getLpaByUid(string $uid, Request $request): array
     {
+        $preferHeader = in_array($uid, $this->preferLpas) ? // we need this to test LPA responses
+            ['Prefer' => sprintf("example=%s", $uid)] : // only while using the sirius mock
+            [];
```
gregtyler, line 118: I don't feel very comfortable about adding code related to mocking things locally to our production codebase.

service-front/module/Application/src/Helpers/LpaFormHelper.php
spinhalf-com: The problem is that the unit tests then fail if I make them private

service-front/module/Application/src/Helpers/LpaFormHelper.php
spinhalf-com: Additional data is when an LPA turns out to have a name or address mismatch, so those details need to be displayed alongside the details for the CP that's identified in the flow. Good idea re the DTO, I'll update it to include that

service-front/module/Application/src/Services/SiriusApiService.php
spinhalf-com: It would need to be removed once the sirius mock is replaced with the actual service. It's needed before that, otherwise there's no way to use the system unless the sirius mock returns behaviourally reliable data

service-front/module/Application/src/Helpers/LpaFormHelper.php
```diff
@@ -0,0 +1,227 @@
...
+
+    private function getDonorNameFromSiriusResponse(array $siriusCheck): string
+    {
+        try {
+            return $siriusCheck['opg.poas.sirius']['donor']['firstname'] .
+                " " .
+                $siriusCheck['opg.poas.sirius']['donor']['surname'];
+        } catch (\Exception $exception) {
+            return $exception->getMessage();
+        }
```
gregtyler, line 109: Similarly this can be cleaned up whilst avoiding warnings:
```suggestion
        return implode(' ', [
            $siriusCheck['opg.poas.sirius']['donor']['firstname'] ?? '',
            $siriusCheck['opg.poas.sirius']['donor']['surname'] ?? '',
        ]);
```

service-front/module/Application/src/Helpers/LpaFormHelper.php
```diff
@@ -0,0 +1,220 @@
...
+        try {
+            foreach ($detailsData['lpas'] as $existingLpa) {
+                if ($lpa == $existingLpa) {
+                    return false;
+                }
+            }
+            return true;
+        } catch (\Exception $exception) {
+            return false;
+        }
```
gregtyler, line 218: Sorry, just spotted that this also has an unnecessary try..catch statement (since there's nothing in the `try` that could throw an exception)
```suggestion
        foreach ($detailsData['lpas'] as $existingLpa) {
            if ($lpa == $existingLpa) {
                return false;
            }
        }

        return true;
```

gregtyler (approved):

PR #131, opened 2024-07-03
Use Twig layouts rather than laminas ones

gregtyler:
**WIP:** For this to work, every template needs to be updated to extend a layout.

---

By default laminas renders your template to `$content`, then renders your layout passing in `$content` as a variable. This is quite limiting as it means we can't use Twig features like blocks.

This PR replaces the laminas template with a complete passthrough (`plain.twig`) and insteads uses Twig's [extends tag](https://twig.symfony.com/doc/3.x/tags/extends.html) to do the layout rendering inside Twig.

This allows us to use blocks, and specifically makes it easy to pair the `<h1>` and `<title>` tags.

It also reduces our dependency on Laminas, as the templates are technically standalone and renderable without the framework.

### DIFF
--- a/service-front/module/Application/config/module.config.php
+++ b/service-front/module/Application/config/module.config.php
@@ -435,6 +435,7 @@ return [
         'doctype'                  => 'HTML5',
         'not_found_template'       => 'error/404',
         'exception_template'       => 'error/index',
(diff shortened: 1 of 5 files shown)


gregtyler:
Done in #142

PR #135, opened 2024-07-04
ID-186-Create-Yoti-session-in-counter-service-journey

fazemiah-moj:
## Purpose

This PR creates the backend API service that integrates with Yoti's APIs to create a user session

This builds on the donar details via CaseData to setup the user profile and address details (currently just in UK formats)

It also replaces our version of the OpenAPI spec ( based on Yoti's official docs) with their versions, this required some amendments also as there were some inaccuracies from API's actual behaviour. 

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -56,22 +64,37 @@ public function findPostOfficeAction(): JsonModel
...
+
+        if (! $uuid) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(['error' => 'Missing uuid']);
         }
-        $this->getResponse()->setStatusCode(Response::STATUS_CODE_204);
-        return new JsonModel(null);
+        $sessionUuid = strval(Uuid::uuid4());
+        $caseData = $this->dataQuery->getCaseByUUID($uuid);
+        $sessionData = $this->sessionConfig->build($caseData, $sessionUuid);
```
gregtyler, line 77: `$authToken` is a clearer description of what the UID is for
```suggestion
        $authToken = strval(Uuid::uuid4());
        $caseData = $this->dataQuery->getCaseByUUID($uuid);
        $sessionData = $this->sessionConfig->build($caseData, $authToken);
```

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -56,22 +64,37 @@ public function findPostOfficeAction(): JsonModel
...
+
+        try {
+            $result = $this->yotiService->createSession($sessionData);
+        } catch (YotiException $e) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(new Problem(
+                'Problem sending request',
+                extra: ['errors' => $e->getMessage()],
+            ));
+        } catch (YotiApiException $apiException) {
```
gregtyler, line 87: What's the difference between YotiException and YotiApiException? They both seem to be thrown as a result of errors from the Yoti API, just from different methods?

I'm also not sure the response codes are right here: `YotiException` will be thrown if the PEM file is invalid, which should result in a 500 error (because the server's misconfigured), not a 400 error.

I feel like `YotiAuthException` would be a good start for a more descriptive class, maybe there are other common more-specific cases too?

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -65,18 +71,48 @@ public function postOfficeBranch(string $postCode): array
...
      * @return array
      * Create a IBV session with applicant data and requirements
+     * is the endpoint there meant to be relative or a full path?
+     * @throws YotiException
+     * @throws YotiApiException
      */
     public function createSession(array $sessionData): array
     {
-        //need to use the RequestSigner for the signature here that is on another branch
+        $sdkId = 'c4321972-7a50-4644-a7cf-cc130c571f59'; //new AwsSecret('yoti/sdk-client-id');
```
gregtyler, line 80: This should be pulled from secrets manager, not hard-coded.

service-api/docker-compose.env
```diff
@@ -9,7 +9,7 @@ MOCK_DRIVINGLICENSE_API=true
 MOCK_KBV_API=true
 MOCK_NINO_API=true
 MOCK_PASSPORT_API=true
-NOTIFICATION_URL=localhost
+NOTIFICATION_URL=https://localhost.com/notify
```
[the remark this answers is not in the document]
fazemiah-moj: yes, that makes more sense

service-api/module/Application/src/Controller/YotiController.php
fazemiah-moj: Think I was using YotiException for pem file/signing issues and then created YotiApiException as a way to differentiate any errors potentially thrown my Yoti itself. 

Maybe YotiAuthException does sound more descriptive to what it's going. Can condense this into one exception with a 500?

gregtyler: I reckon these are three good exceptions to start with:

- YotiAuthException: for failures in the auth/signing process
- YotiClientException: for networking errors (like the Yoti server being unavailable)
- YotiException: for all other unexpected errors (like `/create-session` fails even though we think we gave it correct info)

However, all the handling routes are basically the same (return 500) so it might be worth making YotiAuthException and YotiClientException extend YotiException, so you can just do `catch (YotiException $e)` rather than having to handle all three of them separately.

fazemiah-moj: Sounds good, thanks Greg :)

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -56,22 +63,30 @@ public function findPostOfficeAction(): JsonModel
...
+        $authToken = strval(Uuid::uuid4());
+        $caseData = $this->dataQuery->getCaseByUUID($uuid);
+        $sessionData = $this->sessionConfig->build($caseData, $authToken);
+
+        try {
+            $result = $this->yotiService->createSession($sessionData);
+        } catch (YotiException $e) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_500);
+            return new JsonModel(new Problem(
+                'Problem with request',
```
gregtyler, line 83: Probably better to be more explicit here. If something goes wrong, we're likely to end up with this text as the "culprit" so it's helpful if we can immediately identify that it's a Yoti problem (speaking from experience of trying to debug a problem in Sirius where we just knew "an unexpected error occurred" 😄)
```suggestion
                'Problem requesting Yoti API',
```

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -65,21 +68,49 @@ public function postOfficeBranch(string $postCode): array
...
+            $requestSignature = RequestSigner::generateSignature(
+                '/sessions?sdkId=' . $sdkId . '&nonce=' . $nonce . '&timestamp=' . $timestamp,
+                'POST',
+                new AwsSecret('yoti/certificate'),
+                $body
+            );
+        } catch (Http\Exception\YotiAuthException $e) {
+            $this->logger->error('Yoti Authentication problem for session-create [' . $e->getMessage() . '] ', [
+                'data' => [ 'sessionBody' => $body]
+            ]);
```
gregtyler, line 92: Sorry, previous comment was unclear here: we still need to throw an error to stop it trying to make a request without a signature. I think it's not worth try..catching here and just letting the error bubble up the function above.

Also a general logging point: `$body` contains personal information (name, address, DOB). We try to avoid logging personal info because it becomes a potential security risk.

fazemiah-moj: Ah good point

gregtyler (approved):

PR #143, opened 2024-07-10
Feature/id 215 post office details match page

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-front/module/Application/src/Controller/DonorFlowController.php
+++ b/service-front/module/Application/src/Controller/DonorFlowController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Application\Controller;
 
 use Application\Contracts\OpgApiServiceInterface;
+use Application\Enums\IdMethod;
 use Application\Forms\DrivingLicenceNumber;
 use Application\Forms\NationalInsuranceNumber;
 use Application\Forms\PassportDate;
@@ -56,7 +57,7 @@ class DonorFlowController extends AbstractActionController
                     $view->setVariables($formProcessorResponseDto->getVariables());
                 } else {
                     $this->opgApiService->updateIdMethod($uuid, $formData['id_method']);
-                    if ($formData['id_method'] == 'po') {
+                    if ($formData['id_method'] == IdMethod::PostOffice) {
                         return $this->redirect()->toRoute("root/post_office_documents", ['uuid' => $uuid]);
                     } else {
                         return $this->redirect()->toRoute("root/donor_details_match_check", ['uuid' => $uuid]);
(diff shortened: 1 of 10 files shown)


service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -55,7 +55,11 @@ public function howWillDonorConfirmAction(): ViewModel|Response
                     $view->setVariables($formProcessorResponseDto->getVariables());
                 } else {
                     $this->opgApiService->updateIdMethod($uuid, $formData['id_method']);
-                    return $this->redirect()->toRoute("root/donor_details_match_check", ['uuid' => $uuid]);
+                    if ($formData['id_method'] == 'po') {
```
gregtyler, line 58: It would be good to move these values into an enum (like `IdMethod` in the API) rather than using raw strings.

For example, this isn't in [the list of identity_methods](https://github.com/ministryofjustice/opg-paper-identity/blob/b9394ef9306bcdf91fee221dc8af99fa07b090ae/service-front/module/Application/config/module.config.php#L470-L474) so it's hard to follow where it comes from.

service-front/module/Application/view/application/pages/post_office/donor_details_match_check.twig
```diff
@@ -0,0 +1,48 @@
+{% include 'layout/id_check_banner.twig' with details_data %}
+<div class="govuk-width-container govuk-!-padding-top-5">
+
+        <form method="POST">
+            <h1 class="govuk-heading-xl govuk-!-width-two-thirds">Do the details match the ID document?</h1>
```
gregtyler, line 5: Could you add `govuk-!-width-two-thirds` to the div above, rather than adding it to all of the children?

service-front/module/Application/view/application/pages/post_office/donor_details_match_check.twig
```diff
@@ -0,0 +1,48 @@
...
+            <p class="govuk-body govuk-!-width-two-thirds">
+                The donor’s name and date of birth recorded in Sirius should be identical to
+                how they are presented on the ID document they have chosen to prove their
+                identity with.
+                <br><br>
+                The donor’s address recorded in Sirius should be their home address as this is
+                used for the identity check.
+                <br><br>
+                If there is a mis-match, amend in Sirius as appropriate.
+            </p>
```
gregtyler, line 16: This is a nit, but more semantic to use multiple `<p>` instead of newlines
```suggestion
            <p class="govuk-body govuk-!-width-two-thirds">
                The donor’s name and date of birth recorded in Sirius should be identical to
                how they are presented on the ID document they have chosen to prove their
                identity with.
            </p>
            <p class="govuk-body">
                The donor’s address recorded in Sirius should be their home address as this is
                used for the identity check.
            </p>
            <p class="govuk-body">
                If there is a mis-match, amend in Sirius as appropriate.
            </p>
```

gregtyler (approved):

gregtyler (approved):

PR #147, opened 2024-07-12
Id 227 add lpa type signifier

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist

* [x ] I have added relevant logging with appropriate levels to my code


service-front/web/main.scss
```diff
@@ -53,3 +53,12 @@ $govuk-assets-path: "../assets/";
...
 
+.PW-text-colour {
+  color: #28A197;
+  font-weight: bold;
+}
+
+.PA-text-colour {
+  color: #B58840;
+  font-weight: bold;
+}
```
gregtyler, line 64: Better to use variables for the colours so that we get any changes for free (e.g. if they shift turquoise slightly, it would automatically and consistently here and where we use the same colours in Sirius).

For the classname: it would be good to prefix it with `app` (to avoid conflicts) and move the variation to the end (to use BEM style).

```suggestion
.app-type-text-colour--PW {
  color: govuk-colour("turquoise");
  font-weight: bold;
}

.app-type-text-colour--PA {
  color:  govuk-colour("brown") ;
  font-weight: bold;
}
```

gregtyler (approved):

PR #151, opened 2024-07-15
feature/ID-187-api-generate-return-pdf-letter #151 #minor

fazemiah-moj:
## Purpose

This PR implements the PDF Letter generation via integration with Yoti. The 3 stages to this on Yoti end are mapped to two staged on paper-id end, /prepare-pdf/ and retrieve-pdf

This also refactors the RequestSigner class introduced earlier to remove static function so we can have better test coverage 

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -107,10 +128,58 @@ public function getSessionStatusAction(): JsonModel
...
+        } catch (YotiException $e) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_500);
+            return new JsonModel(new Problem(
+                'Problem retrieving PDF Yoti API',
+                extra: ['errors' => $e->getMessage()],
+            ));
+        }
+        // Write pdf to file
+        $fileName = 'instructions-' . $uuid . '.pdf';
+        $content = file_put_contents($fileName, $data['pdfData']);
```
mattmachell, line 179: Any particular reason to save to the file system? This can leave PII lying about on containers, so is not really advisable and it looks like you are just sending pdfData onward to the response anyway.  If you need to send the file-name onward, you can use a [Content Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header.

fazemiah-moj: This was just so I could see the pdf locally, will remove, as just passing it onwards to caller as response

service-api/module/Application/config/module.config.php
```diff
@@ -237,7 +237,17 @@
                     'route'    => '/counter-service/:uuid/retrieve-letter',
                     'defaults' => [
                         'controller' => Controller\YotiController::class,
-                        'action'     => 'getPDFLetter',
+                        'action'     => 'retrievePDF',
+                    ],
+                ]
+            ],
+            'prepare_pdf_letter' => [
```
mattmachell, line 244: Not sure the counter service internal API needs to map 1-2-1 to the Yoti API, part of the intent of having an internal api is to hide some of the yoti internals. You could wrap the prepare and retrieve into one `/letter` route, also means you avoid the verbs in urls problem..

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -123,12 +116,196 @@ public function retrieveResults(string $sessionId): array
...
+    }
+
+    public function letterConfigPayload(CaseData $caseData, string $requirementId): array
+    {
+        $payload = [];
+
+        $payload["contact_profile"] = [
+            "first_name" => $caseData->firstName,
+            "last_name" => $caseData->lastName,
+            "email" => 'opg-all-team+yoti@digital.justice.gov.uk'
```
mattmachell, line 264: Worth making this email an environment var so it's not hard coed and we can change across environments.

service-api/module/Application/config/module.config.php
gregtyler: I think I would expect one API endpoint to handle:
- Create session
- Generate PDF
- Retrieve PDF and send to Sirius

The front-end shouldn't need to know what's going on behind-the-scenes, just telling the API "start the Yoti journey"

mattmachell: Paper ID will need to call Sirius' API so send the PDF over, rather than Sirius calling Paper ID. As such that probably means the paper id frontend will call generate PDF then immediately call the Sirius documents API to store the letter against the case.

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -63,54 +64,76 @@ public function findPostOfficeAction(): JsonModel
...
+                $notifyAuthToken
+            );
+            //Prepare and generate PDF
+            $this->yotiService->preparePDFLetter($caseData, $nonce, $timestamp, $yotiSessionId);
+            $this->yotiService->retrieveLetterPDF($yotiSessionId, $nonce, $timestamp);
+            //@TODO send pdf from above to sirius when ready
+        } catch (YotiException $e) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_500);
+            return new JsonModel(new Problem(
+                'Problem requesting Yoti API',
```
mattmachell, line 132: Best to be specific about the problem, so "Could not generate and retrieve letter from Yoti API"

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -84,6 +84,14 @@ class CaseData implements JsonSerializable
     #[Annotation\Required(false)]
     public ?string $searchPostcode = null;
 
+    #[Annotation\Required(false)]
+    #[Validator(Uuid::class)]
+    public ?string $sessionId = null;
+
+    #[Annotation\Required(false)]
+    #[Validator(Uuid::class)]
+    public ?string $notifyAuthToken = null;
```
mattmachell, line 93: Notify has a particular meaning in gov, due to the GDS Notify service being called Notify, so best not use that here to avoid confusion. `NotificationsAuthToken` maybe?

docker/yoti/openapi.yml
```diff
@@ -2007,8 +2015,8 @@ components:
           example:
             [
               THANK_YOU_EMAIL_REQUESTED
-              FIRST_BRANCH_VISIT
-              RESOURCE_UPDATE,
+                FIRST_BRANCH_VISIT
```
mattmachell, line 2018: Missing commas?

service-api/module/Application/src/Model/Entity/CaseData.php
gregtyler: Might also be worth putting these inside a Yoti or CounterService model ($caseData->yoti->sessionId) to avoid confusion with our other integrations.

service-api/module/Application/src/Controller/YotiController.php
fazemiah-moj: This does then go onto show more specific messages from the 'errors' part from the exceptions that get bubbled up, so it could broadly be auth issue/signing or specific stages show after the 'Problem requesting Yoti API' message

gregtyler (approved):

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -122,13 +116,194 @@ public function retrieveResults(string $sessionId): array
...
+        $pdf = base64_decode($base64);
+
+        return ["status" => "PDF Created", "pdfData" => $pdf];
+    }
+
+    public function letterConfigPayload(CaseData $caseData, string $requirementId): array
     {
-        return [];
+        $payload = [];
+        $email = (string) getenv("YOTI_LETTER_EMAIL") ?: "opg-all-team+yoti@digital.justice.gov.uk";
```
gregtyler, line 255: It would be good to move this into the `YotiServiceFactory`, so that `YotiService` just has a `notificationEmail` constructor parameter rather than using getenv itself directly.

gregtyler (changes requested):

gregtyler (approved):

PR #152, opened 2024-07-18
Feature/id217 new post office dicuments screen

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -564,4 +564,41 @@ public function updateDobAction(): JsonModel
...
+                'id_method_intl',
+                'M',
+                array_map(fn (mixed $v) => [
+                    'S' => $v
+                ], $data),
+            );
+        } catch (\Exception $exception) {
+            $response['result'] = "Not Updated";
+            $response['error'] = $exception->getMessage();
+            return new JsonModel($response);
```
gregtyler, line 596: This should log the exception, set the response status to 500 and return `new Problem()` rather than  JsonModel.

service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -564,4 +564,41 @@ public function updateDobAction(): JsonModel
...
+        $response = [];
+        $status = Response::STATUS_CODE_200;
+
+        if (! $uuid) {
+            $status = Response::STATUS_CODE_400;
+            $this->getResponse()->setStatusCode($status);
+            $response = [
+                "error" => "Missing UUID"
+            ];
+            return new JsonModel($response);
```
gregtyler, line 581: Errors should return `Problem` so that we have a consistent structure for error messages

```suggestion
            return new Problem("Missing UUID");
```

service-api/module/Application/src/Model/IdMethod.php
```diff
@@ -19,4 +19,8 @@ enum IdMethod: string
     case PostOfficeWithEUDrivingLicense = "po_eud";
     case PostOfficeWithInternationalDrivingLicence = "po_ind";
     case PostOfficeWithNoneOfTheAbove = "po_n";
+    case IntlPassport = 'xpn';
+    case PhotocardDrivingLicence = 'xdln';
+    case NationalIdentityCard = 'xid';
+    case EUId = 'euid';
```
gregtyler, line 25: I'm a bit confused how these fit in with values like `PostOffice` and `PostOfficeWithInternationalPassport`. Maybe this enum needs to be split up into something like "the route the customer has chosen", "the type of ID they're providing" and "the type of country they're providing from (UK/EU/international)"?

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -84,6 +84,9 @@ class CaseData implements JsonSerializable
     #[Annotation\Required(false)]
     public ?string $searchPostcode = null;
 
+    #[Annotation\Required(false)]
+    public ?array $id_method_intl = [];
```
gregtyler, line 88: It's not really clear to me what this new property is for, could it be renamed to have a clearer distinction from `idMethod`?

Should also be camelCase to match the other properties.

service-front/module/Application/config/module.config.php
```diff
@@ -502,5 +522,39 @@
...
+            'Luxembourg',
+            'Malta',
+            'Netherlands',
+            'Poland',
+            'Portugal',
+            'Romania',
+            'Slovakia',
+            'Slovenia',
+            'Spain',
+            'Sweden',
```
gregtyler, line 557: These should be identified and saved as ISO 3166 country codes, then translated into names in the UI. It makes our storage more canonical, and is necessary to pass them to the Yoti API (which uses 3-char ISO-3166).

You should also check these names against [the FCDO list](https://assets.publishing.service.gov.uk/media/65fd8475f1d3a0001132adf4/FCDO_Geographical_Names_Index_March_2024.csv/preview). For example "Republic of Cyprus" should be "Cyprus" and "Czech Republic" should be "Czechia".

service-front/module/Application/view/application/pages/cp/post_office_documents.twig
```diff
@@ -0,0 +1,154 @@
...
+                        <label class="govuk-label govuk-radios__label" for="euid">
+                            EU passport, EU photocard driving licence, EU national identity card (must be current)
+                            <details class="govuk-details govuk-!-margin-top-2 govuk-!-margin-bottom-0">
+                                <summary class="govuk-details__summary">
+                                    <span class="govuk-details__summary-text">
+                                      See which countries are acceptable
+                                    </span>
+                                </summary>
+                                <div class="govuk-details__text">
+                                    Austria, Belgium, Bulgaria, Croatia, Republic of Cyprus, Czech Republic, Denmark, Estonia, Finland, France, Germany, Greece, Hungary, Ireland, Italy, Latvia, Lithuania, Luxembourg, Malta, Netherlands, Poland, Portugal, Romania, Slovakia, Slovenia, Spain and Sweden
```
gregtyler, line 143: This should be pulled from the settings value, not hard-coded.

service-front/package-lock.json
gregtyler, line 1: It looks like you've used `npm` to install this, but we're using `yarn` so this file can be deleted

service-front/web/main.js
```diff
@@ -1,5 +1,26 @@
...
+                const requestedOption = [].filter.call(
+                    this.selectElement.options,
+                    (option) => (option.textContent || option.innerText) === value,
+                )[0];
+                if (requestedOption) {
+                    requestedOption.selected = true;
+                }
+
+                this.selectElement.dispatchEvent(new CustomEvent("confirm"));
+            },
```
gregtyler, line 21: You shouldn't need this block in this repo

service-front/web/main.scss
```diff
@@ -61,3 +61,13 @@ $govuk-assets-path: "../assets/";
...
+
+
+.autocomplete-wrapper {
+  max-width: 20em;
+  margin-bottom: 4rem;
+}
+
+.custom-menu-class {
+  box-shadow: 0 3px 0 0 #fd0, 3px 3px 0 0 #fd0, -3px 3px 0 0 #fd0;
+}
```
gregtyler, line 73: Is this used anywhere? Not sure it's needed

gregtyler (commented):
A reminder to include the ticket (ID-217) and semver tag (#minor) in the PR title or description

service-api/module/Application/src/Model/IdMethod.php
spinhalf-com: Agreed - the id method now stores different types of data at different points in the case journey - eg it stores the post-office ID route as an alternative to the on-the-phone id documents, but then changes this to the eventual id document to be presented at the post office. This needs refactoring, but ideally in a new ticket because it will break other parts of the system

service-front/module/Application/src/Forms/Country.php
```diff
@@ -0,0 +1,28 @@
...
+     * @psalm-suppress PossiblyUnusedProperty
+     */
+    #[Annotation\Validator(NotEmpty::class, ['message' => 'Please select a country'])]
+    public mixed $country;
+
+    /**
+     * @psalm-suppress PossiblyUnusedProperty
+     */
+    #[Annotation\Validator(NotEmpty::class, ['message' => 'Please select a document'])]
+    public mixed $id_method;
```
gregtyler, line 27: It would be good to validate these against a list of accepted values, to avoid saving something that we can't then work with

service-front/module/Application/view/application/pages/cp/choose_country.twig
```diff
@@ -0,0 +1,81 @@
...
+                            <select class="govuk-select" id="country" name="country">
+                                <option value=""></option>
+                                {% for item in countries_data %}
+                                    <option
+                                        value="{{ item }}"
+                                        {% if form.get('country').value == item %}selected{% endif %}
+                                    >
+                                        {{ item }}
+                                    </option>
+                                {%  endfor %}
```
gregtyler, line 47: Should display the country names, but save the ISO code to the DB
```suggestion
                                {% for code, name in countries_data %}
                                    <option
                                        value="{{ code }}"
                                        {% if form.get('country').value == code %}selected{% endif %}
                                    >
                                        {{ name }}
                                    </option>
                                {%  endfor %}
```

gregtyler (approved):

PR #155, opened 2024-07-22
ID-221-Trigger-the-session-creation

fazemiah-moj:
## Purpose

This PR initiates a session creation from the point where the user has confirmed the Post Office route from front end. 

Creation logic has moved to IdentityController so we don't have to make internal requests or forward() which is problematic to test.

Removing endpoint from API also as this doesn't need to be called by anything else.

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Yoti/SessionConfig.php
```diff
@@ -136,7 +136,7 @@ public function addressFormatted(array $address): array
         $addressFormat = [];
         //@TODO determine what address format we are sending, currently no country_iso, assuming all UK for now
         $addressFormat["address_format"] = "1";
-        $addressFormat["building_number"] = substr($address['line1'], 0, 3);
+        $addressFormat["building_number"] = $address['line1'] ? substr($address['line1'], 0, 3) : '1';
```
gregtyler, line 139: This logic doesn't feel right:
- If my address first line is "Eastbourne House", then the building number would be "Eas"?
- If I didn't have a first address line (which is a required field), it would be "1" even though I don't live at Number 1

The first thing is pre-existing but, for this PR, I think if line 1 is empty then it should throw an exception since there's no point starting the ID check process if they don't have an address.

fazemiah-moj: I see that, I think we will need a specific routine/logic function just the work out the line for this "building_number" when we tackle the address issue going forward

gregtyler (approved):

PR #161, opened 2024-07-24
feature/ID-169-ID-Service-API-endpoint-to-handle-Yoti-notification

fazemiah-moj:
## Purpose

This PR handles expected notification data from Yoti. It creates a new secondary index on sessionId 

Due to adding sessionId, this is now a required parameter and needs a default value, which is overwritten when session is actually created.

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/scripts/localstack/init/localstack_init.sh
+++ b/scripts/localstack/init/localstack_init.sh
@@ -8,6 +8,16 @@ awslocal dynamodb create-table \
     AttributeName=id,KeyType=HASH \
   --provisioned-throughput ReadCapacityUnits=1000,WriteCapacityUnits=1000
 
+sleep 5
+
+awslocal dynamodb update-table \
+    --table-name identity-verify \
+    --attribute-definitions AttributeName=sessionId,AttributeType=S \
+    --global-secondary-index-updates \
+        "[{\"Create\":{\"IndexName\": \"sessionId-index\",\"KeySchema\":[{\"AttributeName\":\"sessionId\",\"KeyType\":\"HASH\"}], \
+        \"ProvisionedThroughput\": {\"ReadCapacityUnits\": 20, \"WriteCapacityUnits\": 10 }, \
+        \"Projection\":{\"ProjectionType\":\"ALL\"}}}]"
+
 awslocal secretsmanager create-secret --name local/paper-identity/yoti/certificate \
     --description "PEM certificate for authentication with Yoti API" \
     --secret-string "empty"
--- a/scripts/localstack/init/localstack_init.sh
+++ b/scripts/localstack/init/localstack_init.sh
@@ -12,9 +12,9 @@ sleep 5
 
 awslocal dynamodb update-table \
     --table-name identity-verify \
-    --attribute-definitions AttributeName=sessionId,AttributeType=S \
+    --attribute-definitions AttributeName=yotiSessionId,AttributeType=S \
     --global-secondary-index-updates \
-        "[{\"Create\":{\"IndexName\": \"sessionId-index\",\"KeySchema\":[{\"AttributeName\":\"sessionId\",\"KeyType\":\"HASH\"}], \
+        "[{\"Create\":{\"IndexName\": \"yotiSessionId-index\",\"KeySchema\":[{\"AttributeName\":\"yotiSessionId\",\"KeyType\":\"HASH\"}], \
         \"ProvisionedThroughput\": {\"ReadCapacityUnits\": 20, \"WriteCapacityUnits\": 10 }, \
         \"Projection\":{\"ProjectionType\":\"ALL\"}}}]"
 
--- a/service-api/module/Application/src/Controller/YotiController.php
+++ b/service-api/module/Application/src/Controller/YotiController.php
@@ -6,7 +6,6 @@ namespace Application\Controller;
 
 use Application\Fixtures\DataImportHandler;
 use Application\Fixtures\DataQueryHandler;
-use Application\Model\Entity\CaseData;
 use Application\Model\Entity\Problem;
 use Application\Yoti\Http\Exception\YotiException;
 use Application\Yoti\SessionConfig;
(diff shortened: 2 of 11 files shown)


service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -78,6 +78,9 @@ class CaseData implements JsonSerializable
     #[Annotation\Required(false)]
     public ?string $searchPostcode = null;
 
+    #[Annotation\Required(false)]
+    #[Annotation\Validator(Uuid::class)]
+    public string $sessionId = '7871cae1-6712-4d3f-9111-1c1e38ceb222';
```
gregtyler, line 83: Can we change the name of this field to something that specifically says it's for counter service/Yoti? To avoid future confusion about what session it's referring to.

Also, this shouldn't have a default value

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -82,4 +81,52 @@ public function getSessionStatusAction(): JsonModel
...
+        }
+        $counterServiceMap["notificationState"] = $data['topic'];
+        $this->dataImportHandler->updateCaseData(
+            $caseData->id,
+            'counterService',
+            'M',
+            array_map(fn (mixed $v) => [
+                'S' => $v
+            ], $counterServiceMap),
+        );
```
gregtyler, line 127: Rather than overwriting the whole `counterService` object, could you just update `counterService.notificationState`?
```suggestion
        $this->dataImportHandler->updateCaseData(
            $caseData->id,
            'counterService.notificationState',
            'S',
            $data['topic'],
        );
```

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -82,4 +81,52 @@ public function getSessionStatusAction(): JsonModel
...
+        }
+
+        if (! in_array($data['topic'], ['first_branch_visit', 'session_completion'])) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(new Problem('Invalid type'));
+        }
+
+        $sessionId = filter_var($data['session_id'], FILTER_SANITIZE_SPECIAL_CHARS);
+
+        $caseData = $this->dataQuery->queryByYotiSessionId($sessionId);
```
gregtyler, line 103: This wasn't in the ticket because we ended up doing things in a different order, but we need to authenticate the auth token too by checking that the `Authorization` header matches `$caseData->counterService->notificationsAuthToken`.

NB: the Authorization header will be in the format `Bearer <TOKEN>`, so you need to strip the Bearer bit off first

service-api/module/Application/src/Model/Entity/CaseData.php
fazemiah-moj: Will change name. WRT default value, I needed to add this in as by the point a case entry is first created, we will not have a sessionId, but because I needed to add sessionId as an index, it always needs a value on record creation.

I could not find any work around to this yesterday, will have a think abit more what else we can do.

service-api/module/Application/src/Controller/YotiController.php
fazemiah-moj: Doing this on the follow on ticket ID-170 right now :)

gregtyler: 🤦🏽 Missed that one, thanks

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -82,4 +81,55 @@ public function getSessionStatusAction(): JsonModel
...
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(new Problem('Missing required parameters'));
+        }
+
+        if (! in_array($data['topic'], ['first_branch_visit', 'session_completion'])) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(new Problem('Invalid type'));
+        }
+
+        $sessionId = filter_var($data['session_id'], FILTER_SANITIZE_SPECIAL_CHARS);
```
gregtyler, line 111: It would be safer to do a regex check that this is a UUID

service-api/module/Application/src/Fixtures/DataImportHandler.php
```diff
@@ -92,4 +93,67 @@ public function updateCaseData(string $uuid, string $attrName, string $attrType,
             ]);
         }
     }
+    /**
+     * Specifically for use cases to update map attributes passing in counterService.deadline etc
+     * @psalm-suppress ArgumentTypeCoercion
+     */
+    public function updateCaseChildAttribute(string $uuid, string $attrName, string $attrType, mixed $attrValue): void
```
gregtyler, line 100: This method needs unit tests

service-api/module/Application/src/Fixtures/DataQueryHandler.php
```diff
@@ -49,6 +49,23 @@ public function getCaseByUUID(string $uuid): ?CaseData
         return $arr ? CaseData::fromArray($arr) : null;
     }
 
+    public function queryByYotiSessionId(string $sessionId): ?CaseData
```
gregtyler, line 52: This method needs unit tests

scripts/localstack/init/localstack_init.sh
```diff
@@ -8,6 +8,16 @@ awslocal dynamodb create-table \
...
 
+sleep 5
+
+awslocal dynamodb update-table \
+    --table-name identity-verify \
+    --attribute-definitions AttributeName=yotiSessionId,AttributeType=S \
+    --global-secondary-index-updates \
+        "[{\"Create\":{\"IndexName\": \"yotiSessionId-index\",\"KeySchema\":[{\"AttributeName\":\"yotiSessionId\",\"KeyType\":\"HASH\"}], \
+        \"ProvisionedThroughput\": {\"ReadCapacityUnits\": 20, \"WriteCapacityUnits\": 10 }, \
+        \"Projection\":{\"ProjectionType\":\"ALL\"}}}]"
```
gregtyler, line 19: I've created [a PR to add this to the real infra](https://github.com/ministryofjustice/opg-sirius-infrastructure/pull/2196)

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -119,11 +123,10 @@ public static function fromArray(mixed $data): self
      *     documentComplete: bool,
      *     alternateAddress?: string[],
      *     searchPostcode?: string,
-     *     idMethod?: string,
+     *     yotiSessionId?: string,
      *     counterService?: string[],
-     *     kbvQuestions?: string[]
```
gregtyler, line 124: Why has `kbvQuestions` been removed here? It's still set (line 156)

gregtyler: Discussed on Slack: because there are duplicates (another one needs to be removed on either line 122 or 128)

gregtyler (approved):

PR #173, opened 2024-08-05
ID-179-Create-counter-service-status-check-service

fazemiah-moj:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Yoti/SessionStatusService.php
```diff
@@ -0,0 +1,87 @@
...
+            $response = $this->yotiService->retrieveResults($caseData->yotiSessionId, $nonce, $timestamp);
+            $state = $response['results']['state'];
+            $finalResult = $this->evaluateFinalResult($response['results']['checks']);
+
+            $this->dataImportHandler->updateCaseChildAttribute(
+                $caseData->id,
+                'counterService.state',
+                'S',
+                $state,
+            );
```
gregtyler, line 58: I've just noticed that `updateCaseChildAttribute` (and `updateCaseData`) don't update the connected `CaseData` object, which means you end up with two sources of truth: what's in DynamoDB and what's in memory (i.e. if you then did `echo $caseData->counterService->state;`, it wouldn't be correct).

Probably something for a subsequent PR, but it would be good to fix that. I wonder if, instead of setting individual attributes, we could just re-put the whole object, like:
```php
$caseData->counterService->state = $state;
$this->dynamoDbClient->putItem(['TableName' => ..., 'Item' => $caseData]);
```

service-api/module/Application/config/module.config.php
```diff
@@ -399,6 +401,10 @@
                 $tableName,
                 $serviceLocator->get(LoggerInterface::class)
             ),
+            SessionStatusService::class => fn(ServiceLocatorInterface $serviceLocator) => new SessionStatusService(
+                $serviceLocator->get(YotiServiceInterface::class),
+                $serviceLocator->get(DataImportHandler::class)
+            ),
```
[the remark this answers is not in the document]
fazemiah-moj: ah yes, the auto-wiring did not seem to work here without with the following error: `{
    "status": 500,
    "type": "UnexpectedError",
    "title": "An unexpected error occurred",
    "detail": "Unable to create controller \"Application\\Controller\\YotiController\"; unable to resolve parameter \"sessionService\" using type hint \"Application\\Yoti\\SessionStatusService\"",
    "exception": "Laminas\\ServiceManager\\Exception\\ServiceNotFoundException"
}`

Alternatively, could add this as an invokable factory explicitly with `SessionStatusService::class => InvokableFactory::class`

but that requires creating a factory class for SessionStatus that doesnt have much else in way to configure?

service-api/module/Application/src/Yoti/SessionStatusService.php
```diff
@@ -0,0 +1,87 @@
...
+
+class SessionStatusService
+{
+    public function __construct(
+        public readonly YotiService $yotiService,
+        public readonly DataImportHandler $dataImportHandler
+    ) {
+    }
+
+    public function getSessionStatus(CaseData $caseData): string|array|CounterService
```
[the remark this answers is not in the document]
fazemiah-moj: I thought the whole point of this class would be to prevent unnecessary calls to Yoti if we can derive current status by checking first with notification updates from db. So it would need to return something if the controller was to subsequently use that to fetch the actual results?

service-api/module/Application/src/Yoti/SessionStatusService.php
fazemiah-moj: Oh I missed that one, if I add `            $caseData->counterService->state = $state;
            $caseData->counterService->result = $finalResult;` this could keep them in synch for this issue? Happy to create another ticket to address other areas where the object might be out of line?

With the putItem use above I seem to remember looking into that and what I got was it would overwrite the entire Item, can look into it again

service-api/module/Application/config/module.config.php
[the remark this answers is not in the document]
gregtyler: Sorry I've since spotted that laminas-di has never been installed so there's no autowiring. The way it is is fine for now, though we should probably add autowiring at some point

service-api/module/Application/src/Yoti/SessionStatusService.php
gregtyler: PutItem _will_ overwrite the whole item, but that's somewhat intentional because you already have the whole item in memory (CaseData).

service-api/module/Application/src/Yoti/SessionStatusService.php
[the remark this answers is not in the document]
gregtyler: If you used `PutItem` for the updates, the controller could just read from `CaseData` since it will have the latest information on it. Otherwise I think it should probably return `CounterService` since that contains the data points the controller would need.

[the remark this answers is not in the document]
fazemiah-moj: Ah ok that makes more sense. Will just add logs for any errors in this case :)

service-api/module/Application/src/Controller/YotiController.php
```diff
@@ -72,12 +73,23 @@ public function getSessionStatusAction(): JsonModel
...
-        $session = $this->yotiService->retrieveResults($sessionId);
+
+        $caseData = $this->dataQuery->getCaseByUUID($uuid);
+        if (! $caseData) {
+            $this->getResponse()->setStatusCode(Response::STATUS_CODE_400);
+            return new JsonModel(['error' => 'Case not found']);
+        }
+
+        $sessionId = $caseData->yotiSessionId;
+        if ($sessionId === '00000000-0000-0000-0000-000000000000') {
```
mattmachell, line 84: I'm curious why this particular check exists.

fazemiah-moj: Due to dynamodb quarks, we need a value for an indexed attribute on record creation, so need to check against default value as a false positive

service-api/module/Application/test/Controller/YotiControllerTest.php
```diff
@@ -10,19 +10,22 @@
...
 use Application\Yoti\SessionConfig;
+use Application\Yoti\SessionStatusService;
 use Application\Yoti\YotiService;
 use Application\Yoti\YotiServiceInterface;
 use ApplicationTest\TestCase;
 use Laminas\Http\Headers;
 use Laminas\Http\Request as HttpRequest;
 use Laminas\Http\Response;
 use Laminas\Stdlib\ArrayUtils;
+use phpmock\Mock;
```
gregtyler, line 21: I don't think this is used?

gregtyler (approved):

PR #176, opened 2024-08-06
Feature/id 241 cp flow content changes#minor

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-front/module/Application/config/module.config.php
```diff
@@ -550,33 +570,734 @@
...
+            'NLD' => 'Netherlands',
+            'POL' => 'Poland',
+            'PRT' => 'Portugal',
+            'ROU' => 'Romania',
+            'SKV' => 'Slovakia',
+            'SVN' => 'Slovenia',
+            'ESP' => 'Spain',
+            'SWE' => 'Sweden',
+        ],
+        "supported_countries_documents" => [
```
gregtyler, line 601: Potentially work for a next ticket: but I think we should be fetching this from the Yoti API rather than hard-coding it, in case they change what documents they accept.

service-front/module/Application/src/Helpers/LocalisationHelper.php
```diff
@@ -0,0 +1,73 @@
...
+                $descString .= $word . " ";
+            }
+        }
+        return substr($descString, 0, strlen($descString) - 1);
+    }
+
+    private function getConfig(): array
+    {
+        return include __DIR__ . './../../config/module.config.php';
+    }
```
gregtyler, line 72: You should pass config in through a Factory class, not reference the file path directly. [Example in Sirius](https://github.com/ministryofjustice/opg-sirius/blob/74c059b37ce54bc2b6883ed491b818082dbcae16/back-end/module/Application/src/Authentication/JwtBuilderFactory.php#L9)

service-front/module/Application/src/Helpers/LocalisationHelper.php
```diff
@@ -0,0 +1,73 @@
...
+
+            $documents['supported_documents'][$key] = array_merge(
+                $documents['supported_documents'][$key],
+                ['display_text' => $string]
+            );
+        }
+        return $documents;
+    }
+
+    public function parseWord(string $word): string
```
gregtyler, line 52: It would be much more flexible to just have a mapping from type of document to display name:
```php
$display = [
  'passport' => 'Passport',
  ...
];
```
You could make it [a filter](https://twig.symfony.com/doc/3.x/advanced.html#filters) in TwigExtension, rather than in the controller.

service-front/module/Application/src/Services/OpgApiService.php
```diff
@@ -320,8 +320,19 @@ public function updateIdMethodWithCountry(string $uuid, array $data): array
...
+        $response = $this->makeApiRequest('/identity/details?uuid=' . $uuid);
+        $methodData = [];
+
+        if (array_key_exists('idMethodIncludingNation', $response)) {
+            $methodData = $response['idMethodIncludingNation'];
+        }
+
+        foreach ($data as $key => $value) {
+            $methodData[$key] = $value;
+        }
```
gregtyler, line 332: I don't really understanding what's going on here. It looks like it's fetching `idMethodIncludingNation` so that it can then set `idMethodIncludingNation`?

service-front/module/Application/view/application/pages/cp/choose_country.twig
```diff
@@ -31,45 +31,22 @@
...
+                    <div class="govuk-form-group" id="choose-country">
+                        <label class="govuk-label" for="sort">
 
-                            </label>
-                            <select class="govuk-select" id="country" name="country">
-                                <option value=""></option>
-                                {% for code, name in countries_data %}
-                                    <option
+                        </label>
+                        <select class="govuk-select" id="country" name="country">
```
gregtyler, line 38: This label should have content, and should be `for` an element that exists on the page ("country"?)

service-front/module/Application/src/Controller/CPFlowController.php
```diff
@@ -618,4 +621,41 @@ public function chooseCountryAction(): ViewModel|Response
...
+        }
+        $view->setVariable('form', $form);
+        $view->setVariable('options_data', $idOptionsData);
+        $view->setVariable('countries_data', $idCountriesData);
+        $view->setVariable('countryName', $idCountriesData[
+            $detailsData['idMethodIncludingNation']['country']
+        ]);
+        $view->setVariable('details_data', $detailsData);
+        $view->setVariable('supported_docs', $docs['supported_documents']);
+        $view->setVariable('uuid', $uuid);
```
gregtyler, line 657: You can use `setVariables` to set multiple variables at once

service-front/module/Application/test/Controller/CPFlowControllerTest.php
```diff
@@ -269,4 +274,23 @@ public function testPostOfficeCountriesPage(): void
         $this->assertControllerClass('CpFlowController');
         $this->assertMatchedRouteName('root/cp_choose_country');
     }
+
+    public function testPostOfficeCountriesIdPage(): void
```
gregtyler, line 278: The only thing this test really checks for is that `getDetailsData` is called. It should also cover the logic in the controller code: what's displayed on the page, what happens on submission, validation, saving the data to the API (or those things should be mocked out and tested in smaller units)

service-front/module/Application/view/application/pages/cp/choose_country_Id.twig
```diff
@@ -0,0 +1,50 @@
...
+{% block content %}
+    {% include 'layout/id_check_banner.twig' with details_data %}
+    <div class="govuk-width-container govuk-!-padding-top-5">
+        <form method="POST">
+            <div class="govuk-form-group">
+                <fieldset class="govuk-fieldset" role="group" aria-describedby="passport-issued-hint">
+                    <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+                        <h1 class="govuk-fieldset__heading">
+                            {% block title %}
+                                Choose country ID type
```
gregtyler, line 11: "Choose ID type"? "Choose document"?

service-front/module/Application/view/application/pages/cp/choose_country_Id.twig
```diff
@@ -0,0 +1,50 @@
+{% extends "layout/layout" %}
+{% block content %}
+    {% include 'layout/id_check_banner.twig' with details_data %}
+    <div class="govuk-width-container govuk-!-padding-top-5">
+        <form method="POST">
+            <div class="govuk-form-group">
+                <fieldset class="govuk-fieldset" role="group" aria-describedby="passport-issued-hint">
```
gregtyler, line 7: This aria-describedby attribute refers to a element that doesn't exist.

You should consider adding some [Twig macros](https://twig.symfony.com/doc/3.x/tags/macro.html) for common blocks of HTML (e.g. fieldsets) so you don't have to hand-crank this sort of thing every time.

service-front/module/Application/view/application/pages/post_office/choose_country_id.twig
```diff
@@ -0,0 +1,50 @@
+{% extends "layout/layout" %}
+{% block content %}
+    {% include 'layout/id_check_banner.twig' with details_data %}
+    <div class="govuk-width-container govuk-!-padding-top-5">
+        <form method="POST">
+            <div class="govuk-form-group">
+                <fieldset class="govuk-fieldset" role="group" aria-describedby="passport-issued-hint">
```
gregtyler, line 7: Same issue with aria-describedby here

service-front/module/Application/src/Controller/CPFlowController.php
```diff
@@ -618,4 +621,41 @@ public function chooseCountryAction(): ViewModel|Response
...
+        $form = (new AttributeBuilder())->createForm(CountryDocument::class);
+        $view->setVariable('form', $form);
+
+        if ($this->getRequest()->isPost()) {
+            $form->setData($this->getRequest()->getPost());
+            $formData = $this->getRequest()->getPost()->toArray();
+
+            if ($form->isValid()) {
+                $responseData = $this->opgApiService->updateIdMethodWithCountry($uuid, $formData);
+                if ($responseData['result'] === 'Updated') {
```
gregtyler, line 644: We should just be checking that the response code is 2xx, rather than passing around random case-sensitive strings

service-front/module/Application/config/module.config.php
spinhalf-com: It's complicated because OPG is putting additional rules on the Yoti document list according to yesterday's planning. This static array will need to be updated for that anyway.

service-front/module/Application/src/Services/OpgApiService.php
spinhalf-com: idMethodIncludingNation is an array containing two pieces of data, the country and the document type. At this point it will have the country defined but not the document type, this is updating the array and then sending it back to the API

gregtyler: So if `idMethodIncludingNation` is set in the form, line 327 is irrelevant because it will then get overwritten on line 331.

Is the point that if `idMethodIncludingNation` is already set, but it _isn't_ set on the form, it would be replaced with blank, so line 327 is making sure it's not overwritten?

service-front/module/Application/test/Controller/CPFlowControllerTest.php
spinhalf-com: All the controller function tests do this: they simply check that the route loads the page without errors and nothing more. I understood that we were going to use cypress for testing content and behaviour?

service-front/module/Application/config/module.config.php
gregtyler: In that case I believe we need to:
- Include passports from non-EU countries in this
- Remove "TRAVEL_DOCUMENT" since that's not on our list of accepted ID

It's probably also worth removing `is_strictly_latin` and `date_from` since we're not using them.

We should also ask Yoti how we can be notified when their supported documents change, to avoid sending people to the Post Office with something that won't be supported.

service-front/module/Application/src/Helpers/LocalisationHelper.php
```diff
@@ -0,0 +1,73 @@
...
+        'DRIVING_LICENCE' => 'Driving licence',
+        'NATIONAL_ID' => 'National ID',
+        'RESIDENCE_PERMIT' => 'Residence permit',
+        'TRAVEL_DOCUMENT' => 'Travel document',
+    ];
+
+    /**
+     * @throws LocalisationException
+     */
+    public function getInternationalSupportedDocuments(array $detailsData): array
```
gregtyler, line 26: To avoid a dependency on the shape of the case data, could this just take `countryCode` as an argument?
```suggestion
    public function getInternationalSupportedDocuments(string $countryCode): array
```

service-front/module/Application/test/Helpers/LocalisationHelperTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+use Application\Helpers\LocalisationHelper;
+use Application\Helpers\LpaFormHelper;
+use Laminas\Form\Annotation\AttributeBuilder;
+use Laminas\Form\FormInterface;
+use Laminas\Stdlib\Parameters;
+use PHPUnit\Framework\TestCase;
+use Application\Forms\LpaReferenceNumber;
+
+class LocalisationHelperTest extends TestCase
```
gregtyler, line 15: This should have a test for `getInternationalSupportedDocuments`, since that's the only real publicly-exposed method

service-front/module/Application/test/Controller/CPFlowControllerTest.php
gregtyler: We try to only use Cypress for happy path smoke tests, because it's so much slower and harder to run/debug than PHPUnit tests (per the [Test Pyramid](https://martinfowler.com/bliki/TestPyramid.html)).

Appreciate we've not been treating tests as a high priority, but as the application gets more complicated we're running into more regressions so we really need to build up our test coverage going forward.

service-front/module/Application/src/Services/OpgApiService.php
spinhalf-com: It's a way of ensuring that if either key is set in the new data but not both, the existing key isn't overwritten with a blank. I know it's a bit inelegant, but it's what worked.

service-front/module/Application/test/Controller/CPFlowControllerTest.php
```diff
@@ -269,4 +274,101 @@ public function testPostOfficeCountriesPage(): void
...
+    public function testPostOfficeCountriesIdPostPage(): void
+    {
+        $mockResponseDataIdDetails = $this->returnOpgResponseData();
+
+        $this
+            ->opgApiServiceMock
+            ->expects(self::once())
+            ->method('getDetailsData')
+            ->with($this->uuid)
+            ->willReturn($mockResponseDataIdDetails);
```
gregtyler, line 360: Should also test the new data is saved:
```php
$this
    ->opgApiServiceMock
    ->expects(self::once())
    ->method('updateIdMethodWithCountry')
    ->with($this->uuid, ['id_method' => 'PASSPORT']);
```
Not least because the shape of `idMethodIncludingNation` doesn't seem to be defined anywhere

gregtyler (approved):

PR #178, opened 2024-08-07
ID-291-donor-CP-bug-post-office-does-not

fazemiah-moj:
## Purpose

This PR fixed a bug with post code attribute naming, which was incorrectly specified in Yoti docks and this build as reference with mocks and tests

It also fixes the different use of counterService->selectedPostOffice is being send from FE as an object rather than simply containing fad_code

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-api/module/Application/src/Yoti/YotiService.php
+++ b/service-api/module/Application/src/Yoti/YotiService.php
@@ -271,9 +271,10 @@ class YotiService implements YotiServiceInterface
         ];
         /** @var CounterService $counterServiceData */
         $counterServiceData = $caseData->counterService;
+        $postOfficeData = json_decode($counterServiceData->selectedPostOffice, true);
         $payload["branch"] = [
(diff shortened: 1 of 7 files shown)


mattmachell (approved):

service-api/module/Application/src/Yoti/YotiService.php
```diff
@@ -271,9 +271,10 @@ public function letterConfigPayload(CaseData $caseData, string $requirementId):
         ];
         /** @var CounterService $counterServiceData */
         $counterServiceData = $caseData->counterService;
+        $postOfficeData = json_decode($counterServiceData->selectedPostOffice, true);
```
gregtyler, line 274: I hadn't spotted that we were stored `selectedPostOffice` as a JSON blob. Not a blocker for this PR but [I've put in a ticket](https://opgtransform.atlassian.net/browse/ID-295) to revisit that

PR #181, opened 2024-08-08
ID-294-documentComplete-not-being-called-on-post-office-Donor-flow

fazemiah-moj:
This PR fixes bug resulting from miscommunication regarding where the front end process should actually call the API to trigger the session creation

This now moves the Yoti process back to YotiController and re-instates its own endpoint so it can be called explicitly. 
And because this is now called explicitly, it circumvents the bug in https://opgtransform.atlassian.net/browse/ID-293 for now

## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-front/module/Application/src/Controller/DonorPostOfficeFlowController.php
```diff
@@ -173,7 +173,8 @@ public function confirmPostOfficeAction(): ViewModel|Response
 
         if (count($this->getRequest()->getPost())) {
             $responseData = $this->opgApiService->confirmSelectedPostOffice($uuid, $deadline);
-            if ($responseData['result'] == 'Updated') {
+            $counterService = $this->opgApiService->createYotiSession($uuid);
+            if ($responseData['result'] == 'Updated' && $counterService['result'] == 'Session created') {
```
gregtyler, line 177: If the request wasn't successful, the API will return a non-200 response and front container will throw an `OpgApiException`.

Therefore there's no point checking the content of the response here because if it's anything else it will already have thrown an exception.

Also, in general, we should be using human-readable strings to pass state between API and frontend. If additional information _were_ needed it should be in a boolean or well-defined enum.

fazemiah-moj: So we can safely remove this whole block ?
`  if ($responseData['result'] == 'Updated' && $counterService['result'] == 'Session created') {
                return $this->redirect()->toRoute('root/what_happens_next', ['uuid' => $uuid]);
            } else {
                $view->setVariable('errors', ['API Error']);
            }`

gregtyler: Sorry, also just noticed: I think this should be called in the POST handler for `findPostOfficeBranchAction`, since that's where the selected post office gets confirmed. This page must be shown after the session has been started, because it displays the deadline.

fazemiah-moj: From what I can see _findPostOfficeBranchAction_ handles the first selection stage of the post office branch, and its confirmed at _confirmPostOfficeAction_ again. Can move it up the chain there but wondering if it could lead to a session creation and user flow backing out of post office route?

gregtyler: Probably best to check with Matthew/Jude what their expectations are. If they don't want the session to start until the second page is submitted then we'll need to add an API endpoint to fetch the _expected_ deadline for display on that page.

fazemiah-moj: Ok for now I've moved it up, since we are still testing. Will check with Matthew/Jude if this would cause problems

mattmachell (approved):

PR #220, opened 2024-09-03
ID-334-Edit-the-donor-details-on-Sirius-page-should-take-user-to-Sirius

fazemiah-moj:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-front/module/Application/src/Controller/DonorFlowController.php
+++ b/service-front/module/Application/src/Controller/DonorFlowController.php
@@ -96,10 +96,13 @@ class DonorFlowController extends AbstractActionController
 
         $view = new ViewModel();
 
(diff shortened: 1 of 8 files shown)


service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -95,11 +95,15 @@ public function donorDetailsMatchCheckAction(): ViewModel
         $detailsData['formatted_dob'] = (new \DateTime($detailsData['dob']))->format("d F Y");
 
         $view = new ViewModel();
+        /** @var string $base */
+        $base = getenv("SIRIUS_BASE_URL");
```
gregtyler, line 99: We don't like using `getenv` in application code because it creates a configuration injection dependency and makes it hard to mock. Instead, call `getenv` in the factory and pass through the value as a constructor variable (e.g. `public function __construct(private readonly string $siriusBaseUrl`)

PR #223, opened 2024-09-04
Generate auth token for KBVService

gregtyler:
## Purpose

Generate auth token to be used for requests to the Experian IIQ API

Fixes ID-257 #minor

## Approach

Implement the "real" KBV Service which:
- Generates an auth token through the Wasp API
- Uses the mock to fetch KBV questions (so that it can continue to function)

The Soap API is split into two parts: **WaspClient** is a configured extension of SoapClient which allows you to send SOAP messages. This shouldn't be used directly as the API is not clearly defined, but stands alone for easier configuration and mocking.

**WaspService** wraps the SOAP operations that we use to provide a clearer inteface. This means that its methods should have minimal parameters and return types, and it should translate those with the SOAP operation definition. WaspService should not contain any business logic, just wrap operations.

## Checklist

  * N/A
* [x] The team have tested these changes
  * N/A


gregtyler:
Confirmed working in a popup env: "Token length 2236"

PR #227, opened 2024-09-05
Feature/id 278 crosscore api php client#major

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/Services/Experian/FraudApi/ExperianCrosscoreFraudApiService.php
```diff
@@ -0,0 +1,172 @@
...
+
+            $responseArray = json_decode($response->getBody()->getContents(), true);
+
+            return new ExperianCrosscoreFraudResponseDTO(
+                $responseArray
+            );
+        } catch (ClientException $clientException) {
+            if ($clientException->getResponse()->getStatusCode() == Response::STATUS_CODE_401) {
+                $this->experianCrosscoreAuthApiService->authenticate();
+                $this->makeRequest($experianCrosscoreFraudRequestDTO);
```
gregtyler, line 82: If the authentication fails reliably (e.g. if we haven't got the right credentials), it will infinitely loop. I'd suggest adding a parameter to `makeRequest` to indicate whether it's a retry or not, so you don't retry the retry

service-api/module/Application/src/Services/Experian/FraudApi/ExperianCrosscoreFraudApiService.php
```diff
@@ -0,0 +1,172 @@
...
+        $personId = $this->makePersonId($experianCrosscoreFraudRequestDTO);
+        $addressDTO = $experianCrosscoreFraudRequestDTO->address();
+
+        $body = [
+            'header' => [
+                "tenantId" => $this->config['tenantId'],
+                "requestType" => "FraudScore",
+                "clientReferenceId" => "$requestUuid-FraudScore-continue",
+                "expRequestId" => $requestUuid,
+                "messageTime" => date("Y-m-d\TH:i:s.000\Z", time()),
```
gregtyler, line 102: `time()` is the default second argument of `date()`
```suggestion
                "messageTime" => date("Y-m-d\TH:i:s.000\Z"),
```

service-api/module/Application/src/Services/Experian/FraudApi/DTO/ExperianCrosscoreFraudResponseDTO.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Services\Experian\FraudApi\DTO;
+
+use Application\Services\Experian\FraudApi\ExperianCrosscoreFraudApiException;
+use phpDocumentor\Reflection\PseudoTypes\Numeric_;
+
+class ExperianCrosscoreFraudResponseDTO
```
gregtyler, line 10: This DTO is returning pretty much everything, but we don't need all this information. It would be good to cut it down to just the bits we actually want, which would also make the tests a lot neater because there wouldn't be as much unused info in there.

service-front/composer.json
```diff
@@ -26,7 +26,8 @@
         "laminas/laminas-session": "^2.16.0",
         "laminas/laminas-skeleton-installer": "^1.3.0",
         "monolog/monolog": "^3.5",
-        "oxcom/zend-twig": "^1.1"
+        "oxcom/zend-twig": "^1.1",
+        "ramsey/uuid": "^4.7"
```
gregtyler, line 30: I don't think this should be in service-front?

service-front/docker-compose.env
```diff
@@ -3,3 +3,5 @@ API_BASE_URI=http://api-web
 SIRIUS_BASE_URL=http://sirius-mock:4010 # http://host.docker.internal:8080 for real Sirius
 SIRIUS_PUBLIC_URL=http://localhost:8080
 APP_DEBUG=1
+EXPERIAN_AUTH_URL=https://uk-api.experian.com/oauth2/experianone/v1/token
+EXPERIAN_BASE_URL=https://uk-api.experian.com
```
gregtyler, line 7: These aren't used by the front-end

service-front/phpunit.xml
```diff
@@ -29,7 +29,7 @@
         <env name="SIRIUS_BASE_URL" value="SIRIUS_BASE_URL" />
         <env name="SIRIUS_PUBLIC_URL" value="SIRIUS_PUBLIC_URL" />
         <env name="DISABLE_AUTH_LISTENER" value="1"/>
-
+        <env name="EXPERIAN_AUTH_URL" value="EXPERIAN_AUTH_URL"/>
```
gregtyler, line 32: This isn't used by the front-end

service-api/module/Application/src/Services/Experian/FraudApi/DTO/ExperianCrosscoreFraudResponseDTO.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Services\Experian\FraudApi\DTO;
+
+use Application\Services\Experian\FraudApi\ExperianCrosscoreFraudApiException;
+use phpDocumentor\Reflection\PseudoTypes\Numeric_;
```
gregtyler, line 8: Doesn't look like this is used

service-api/module/Application/src/Experian/Crosscore/FraudApi/FraudApiService.php
```diff
@@ -0,0 +1,175 @@
...
+            return new ResponseDTO(
+                $responseArray
+            );
+        } catch (ClientException $clientException) {
+            if ($clientException->getResponse()->getStatusCode() == Response::STATUS_CODE_401) {
+                $this->experianCrosscoreAuthApiService->authenticate();
+                $this->makeRequest($experianCrosscoreFraudRequestDTO);
+            } elseif ($this->authCount > 2) {
+                throw $clientException;
+            } else {
```
gregtyler, line 84: Need to check `authCount` in the initial if statement or the first one will always trigger:
```suggestion
            if ($clientException->getResponse()->getStatusCode() == Response::STATUS_CODE_401 && $this->authCount < 2) {
                $this->experianCrosscoreAuthApiService->authenticate();
                $this->makeRequest($experianCrosscoreFraudRequestDTO);
            } else {
```

service-api/module/Application/test/ApplicationTest/Services/Experian/FraudApi/DTO/ExperianCrosscoreFraudResponseDTOTest.php
```diff
@@ -0,0 +1,255 @@
...
+{
+    private ResponseDTO $experianCrosscoreFraudResponseDTO;
+
+    private array $data;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->data = [
```
gregtyler, line 20: You can minimise this to just the bits that we actually use, since everything else is ignored.

There's the same in a few other test files. At a minimum, I think the `clientResponsePayload` and `originalRequestData` bits can be removed everywhere.

service-front/composer.lock
gregtyler, line 1: Changes to composer.json were reverted, the ones here should be taken back too

gregtyler (approved):

PR #265, opened 2024-10-11
Feature/id 300 dependency healthcheck

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-front/module/Application/src/Helpers/DependencyCheck.php
```diff
@@ -17,24 +17,40 @@ public function __construct(private readonly array $statusData)
...
+            $processedData[IdMethod::DrivingLicenseNumber->value] === false &&
+            $processedData[IdMethod::PassportNumber->value] === false &&
+            $processedData[IdMethod::NationalInsuranceNumber->value] === false
+        ) {
+            $processedData['EXPERIAN'] = false;
+            $message = "Online identity verification it not presently available";
+        }
+
+        $this->processedStatus['data'] = $processedData;
+        $this->processedStatus['message'] = $message;
```
gregtyler, line 53: You've got a whole object here, so might as well use its properties rather than stuff everything into an array
```suggestion
        $this->processedStatus = $processedData;
        $this->message = $message;
```

service-front/module/Application/config/module.config.php
```diff
@@ -594,10 +594,10 @@
...
         ],
-        'identity_methods' => [
-            'nin' => 'National Insurance number',
-            'pn' => 'UK Passport (current or expired in the last 18 months)',
-            'dln' => 'UK driving licence (must be current) ',
+        'identity_labels' => [
+            'NATIONAL_INSURANCE_NUMBER' => 'National Insurance number',
+            'PASSPORT' => 'UK Passport (current or expired in the last 18 months)',
+            'DRIVING_LICENCE' => 'UK photocard driving licence (must be current) ',
         ],
```
gregtyler, line 601: I think `DonorFlowController::donorIdCheckAction` needs to be updated since it still refers to identity_methods?

Or potentially that method and this config can be removed.

service-front/module/Application/src/Services/OpgApiService.php
```diff
@@ -366,6 +366,6 @@ public function getServiceAvailability(): array
             throw new OpgApiException('Service availability data missing!');
         }
 
-        return (new DependencyCheck($this->responseData))->getProcessedStatus();
+        return (new DependencyCheck($this->responseData))->toArray();
```
gregtyler, line 369: If you return an array, you're throwing away the data structure and losing any benefit of having an object/properties. Better to return `DependencyCheck` itself so you can be confident what properties are available.
```suggestion
        return new DependencyCheck($this->responseData);
```
Or, if you want to use array shapes instead of an object, make DependencyCheck [an invokable](https://www.php.net/manual/en/language.oop5.magic.php#object.invoke) since it doesn't need to carry state.

gregtyler (approved):

PR #271, opened 2024-10-18
what is vouching page

davidbridgwood:
## Purpose

Adds the "what is vouching" and "what happens next" pages to the Donor journey.

Fixes: [ID-305](https://opgtransform.atlassian.net/browse/ID-305),  [ID-306](https://opgtransform.atlassian.net/browse/ID-306) #minor

## Approach

## Learning

## Checklist

[ID-305]: https://opgtransform.atlassian.net/browse/ID-305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ID-306]: https://opgtransform.atlassian.net/browse/ID-306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docker/sirius/openapi.yml
+++ b/docker/sirius/openapi.yml
@@ -285,11 +285,12 @@ components:
             - success
             - exit
       additionalProperties: false
+    # do i just rename this, or create a new schema??
     YotiPdfSuffixRequest:
       type: object
       required:
         - systemType
-        - pdfSuffix
+        # - pdfSuffix
         - correspondentName
       properties:
         type:
--- a/docker/sirius/openapi.yml
+++ b/docker/sirius/openapi.yml
@@ -143,16 +143,16 @@ paths:
           schema:
             type: integer
             nullable: false
-      operationId: postPostOfficeLetterSuffix
-      summary: Send Yoti Letter PDF data
-      description: Send base64 encoded PDF data for Post Office letter
+      operationId: postCreateDocumentRequest
+      summary: Send letter data
+      description: Send data to create letter
       tags: [ poa ]
       requestBody:
         description: Sirius LPA document data
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/YotiPdfSuffixRequest"
+              $ref: "#/components/schemas/CreateDocumentRequest"
       responses:
         "201":
           description: PDD data has been processed
@@ -285,12 +285,10 @@ components:
             - success
             - exit
       additionalProperties: false
-    # do i just rename this, or create a new schema??
-    YotiPdfSuffixRequest:
+    CreateDocumentRequest:
       type: object
       required:
         - systemType
-        # - pdfSuffix
         - correspondentName
       properties:
         type:
--- a/service-front/module/Application/src/Controller/DonorFlowController.php
+++ b/service-front/module/Application/src/Controller/DonorFlowController.php
@@ -116,13 +116,26 @@ class DonorFlowController extends AbstractActionController
         if ($this->getRequest()->isPost()) {
             $formData = $this->getRequest()->getPost()->toArray();
             if ($formData['confirm_vouching'] == 'yes') {
-                return $this->redirect()->toRoute("root/what_is_vouching", ['uuid' => $uuid]);
+                return $this->redirect()->toRoute("root/vouching_what_happens_next", ['uuid' => $uuid]);
             } else {
                 return $this->redirect()->toRoute("root/how_donor_confirms", ['uuid' => $uuid]);
             }
         }
 
         return $view->setTemplate('application/pages/what_is_vouching');
+    }
+
+    public function vouchingWhatHappensNextAction(): ViewModel|Response
+    {
+        $view = new ViewModel();
+        $uuid = $this->params()->fromRoute("uuid");
+        $detailsData = $this->opgApiService->getDetailsData($uuid);
+        $siriusEditUrl = $this->siriusPublicUrl . '/lpa/frontend/lpa/' . $detailsData["lpas"][0];
+
+        $view->setVariable('details_data', $detailsData);
+        $view->setVariable('sirius_edit_url', $siriusEditUrl);
+
+        return $view->setTemplate('application/pages/vouching_what_happens_next');
     }
 
     public function donorDetailsMatchCheckAction(): ViewModel
--- a/service-front/module/Application/src/Controller/DonorFlowController.php
+++ b/service-front/module/Application/src/Controller/DonorFlowController.php
@@ -85,6 +85,8 @@ class DonorFlowController extends AbstractActionController
                             $data
                         );
                         return $this->redirect()->toRoute("root/post_office_documents", ['uuid' => $uuid]);
+                    } elseif ($formData['id_method'] == IdMethod::OnBehalf->value) {
+                        return $this->redirect()->toRoute("root/what_is_vouching", ['uuid' => $uuid]);
                     } else {
                         $data = [
                             'id_route' => 'TELEPHONE',
@@ -102,6 +104,25 @@ class DonorFlowController extends AbstractActionController
         }
 
         return $view->setTemplate($templates['default']);
+    }
+
+    public function whatIsVouchingAction(): ViewModel|Response
+    {
+        $view = new ViewModel();
+        $uuid = $this->params()->fromRoute("uuid");
+        $detailsData = $this->opgApiService->getDetailsData($uuid);
+        $view->setVariable('details_data', $detailsData);
+
+        if ($this->getRequest()->isPost()) {
+            $formData = $this->getRequest()->getPost()->toArray();
+            if ($formData['confirm_vouching'] == 'yes') {
+                return $this->redirect()->toRoute("root/what_is_vouching", ['uuid' => $uuid]);
+            } else {
+                return $this->redirect()->toRoute("root/how_donor_confirms", ['uuid' => $uuid]);
+            }
+        }
+
+        return $view->setTemplate('application/pages/what_is_vouching');
     }
 
     public function donorDetailsMatchCheckAction(): ViewModel
--- a/service-front/module/Application/src/Controller/DonorFlowController.php
+++ b/service-front/module/Application/src/Controller/DonorFlowController.php
@@ -116,7 +116,10 @@ class DonorFlowController extends AbstractActionController
         if ($this->getRequest()->isPost()) {
             $formData = $this->getRequest()->getPost()->toArray();
             if ($formData['confirm_vouching'] == 'yes') {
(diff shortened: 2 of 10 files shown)


service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -103,6 +105,26 @@ public function howWillDonorConfirmAction(): ViewModel|Response
...
+        $uuid = $this->params()->fromRoute("uuid");
+        $detailsData = $this->opgApiService->getDetailsData($uuid);
+        $view->setVariable('details_data', $detailsData);
+
+        if ($this->getRequest()->isPost()) {
+            $formData = $this->getRequest()->getPost()->toArray();
+            if ($formData['confirm_vouching'] == 'yes') {
+                return $this->redirect()->toRoute("root/what_is_vouching", ['uuid' => $uuid]);
+            }
+            else {
```
gregtyler, line 120:
```suggestion
            } else {
```

service-front/module/Application/test/Controller/DonorFlowControllerTest.php
```diff
@@ -233,6 +233,41 @@ public function testDonorIdMatchPage(): void
...
+        $this->assertControllerClass('DonorFlowController');
+        $this->assertMatchedRouteName('root/what_is_vouching');
+
+        $this->dispatch("/$this->uuid/what-is-vouching", 'POST', [
+            'confirm_vouching' => $confirmVouching,
+         ]);
+
+        $this->assertResponseStatusCode(302);
+        $this->assertRedirectTo(sprintf($expectedRedirect, $this->uuid));
+
```
gregtyler, line 254:
```suggestion
```

service-front/module/Application/src/Services/SiriusApiService.php
```diff
@@ -160,15 +160,20 @@ public function abandonCase(array $data, Request $request): array
...
      * @param array $caseDetails
+     * @param string $systemType
      * @param Request $request
+     * @param string $base64suffix
      * @return array
      * @throws GuzzleException
      */
-    public function sendPostOfficePDf(string $base64suffix, array $caseDetails, Request $request): array
-    {
+    public function sendPdf(
```
davidbridgwood, line 171: i've changed this to be a more generic function that can be used to send both the post-office and vouching letters since theres a significant overlap.
Having said that i'm not 100% sure what the significance of the `$base64suffix` parameter is - does its absence from the vouching letter mean its not a PDF? Does that mean my function name doesn't make sense? Should I actually keep them as 2 functions?

service-front/module/Application/test/Controller/DonorFlowControllerTest.php
```diff
@@ -234,6 +234,78 @@ public function testDonorIdMatchPage(): void
     }
 
 
+    public function testWhatIsVouchingPageOptNo(): void
```
davidbridgwood, line 237: have added 3 separate tests for the what-is-vouching page as it was getting complex using a `dataProvider` but i'm not sure what best practice would be here, should i stick to a single test with a `dataProvider` and conditional logic in the test?

service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -95,15 +97,57 @@ public function howWillDonorConfirmAction(): ViewModel|Response
...
+            if ($formData['confirm_vouching'] == 'yes') {
+                /**
+                 * @psalm-suppress ArgumentTypeCoercion
+                 */
+                $pdf = $this->siriusApiService->sendPdf($detailsData, "DLP-VOUCH-INVITE", $this->request);
+                if ($pdf['status'] === 201) {
+                    return $this->redirect()->toRoute("root/vouching_what_happens_next", ['uuid' => $uuid]);
+                } else {
+                    // what should i do with this error??
+                    $view->setVariable('errors', ['API Error - Could not send letter']);
```
mattmachell, line 128: Does the template you're using handle displaying errors? If so you can let the view rendering take care of it. If not then you may want to throw an exception that the generic handlers will pick up. Also worth logging that the error occured.

service-front/module/Application/view/application/pages/what_is_vouching.twig
```diff
@@ -0,0 +1,63 @@
+{% extends "layout/layout" %}
+
+{% block beforeMain %}
+    {% include 'layout/id_check_banner.twig' with details_data %}
+{% endblock %}
+
+{% block content %}
+    <form method="POST" class="govuk-!-width-two-thirds">
+        <h1 class="govuk-heading-xl">{% block title %}What is Vouching?{% endblock %}</h1>
```
mattmachell, line 9: Non-form elements inside forms can get weird for assistive tech, I'd move the `<form>` to wrap just the form-group with the radios at line 41.

service-front/module/Application/view/application/pages/what_is_vouching.twig
```diff
@@ -0,0 +1,63 @@
...
+                    <li>a member of your family</li>
+                    <li>living at the same address as you</li>
+                </ul>
+            </p>
+        </div>
+
+        <div class="govuk-form-group">
+            <h2 class="govuk-heading-m">Choose someone else to vouch for the donor's identity?</h2>
+            <div class="govuk-radios__item">
+                <input class="govuk-radios__input" id="yes" name="confirm_vouching" type="radio" value="yes">
```
mattmachell, line 44: I'd be wary of using general ids like `yes` and `no` on form elements like this. Scoping it to the form name with a prefix will prevent ID collisions in the rest of the page. Something like `confirm_vouching_yes`.

Also looks like you're missing the fieldset and legend elements this component expects: https://design-system.service.gov.uk/components/radios/

docker/sirius/openapi.yml
```diff
@@ -285,11 +285,12 @@ components:
             - success
             - exit
       additionalProperties: false
+    # do i just rename this, or create a new schema??
     YotiPdfSuffixRequest:
```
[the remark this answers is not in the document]
gregtyler: Yeah, there's nothing specifically Yoti about that endpoint (it's used by other frontends too) so probably worth renaming to something like "CreateDocumentRequest". Similarly the operation ID "postPostOfficeLetterSuffix" could be genericised.

service-front/module/Application/src/Services/SiriusApiService.php
gregtyler: You've handled that spot on: it's an optional parameter of an existing PDF file in base-64 format. If you provide it, Sirius will generate the letter then stitch the `base64suffix` afterwards. In Yoti context, this is us sending a PDF that they've generated with a custom cover letter.

Might be worth renaming that variable to something like `pdfSuffixBase64` to be a bit clearer, and maybe updating the `@param` to give quick explanation of what it's for.

service-front/module/Application/test/Controller/DonorFlowControllerTest.php
gregtyler: No, I think that's good. It's only worth using a dataprovider when you have lots of clearly similar tests, like testing a validation function.

docker/sirius/openapi.yml
```diff
@@ -285,11 +285,12 @@ components:
...
       type: object
       required:
         - systemType
-        - pdfSuffix
+        # - pdfSuffix
```
gregtyler, line 293: You're right this isn't needed, so it can be completely removed:
```suggestion
```

service-front/module/Application/src/Controller/DonorFlowController.php
```diff
@@ -95,15 +97,57 @@ public function howWillDonorConfirmAction(): ViewModel|Response
...
+        $detailsData = $this->opgApiService->getDetailsData($uuid);
+        $view->setVariable('details_data', $detailsData);
+
+        if ($this->getRequest()->isPost()) {
+            $formData = $this->getRequest()->getPost()->toArray();
+            if ($formData['confirm_vouching'] == 'yes') {
+                /**
+                 * @psalm-suppress ArgumentTypeCoercion
+                 */
+                $pdf = $this->siriusApiService->sendPdf($detailsData, "DLP-VOUCH-INVITE", $this->request);
```
gregtyler, line 123: If you use `getRequest` you can avoid needing the Psalm suppression, since it will always return `Laminas\Http\Request`
```suggestion
                $pdf = $this->siriusApiService->sendPdf($detailsData, "DLP-VOUCH-INVITE", $this->getRequest());
```

service-front/module/Application/src/Controller/DonorFlowController.php
gregtyler: Worth testing by breaking the mock (e.g. remove the 201 response from the OpenAPI file), but I think it will already throw an exception on failure and that will be caught by the framework to show an error page.

i.e. you don't need to add any handling for the error condition as this code will never be reached

service-front/module/Application/src/Controller/DonorFlowController.php
gregtyler: Also I'd suggest moving the template IDs to either an enum or class constants, rather than having them in code here, so that you can do something like:
```php
$this->siriusApiService->sendPdf($detailsData, SiriusDocument::VouchInvitation, $this->request);
```
That isolates the Sirius configuration into a centralised file in case it needs to be changed, and makes the intention of the document clearer (not so much a problem here, but I'm guessing "DLP-ID-PO-D" doesn't mean much to most people working Paper ID 😁)

service-front/module/Application/src/Controller/DonorFlowController.php
davidbridgwood: yep - tested removing the 201 response and you get sent to an error page and never see this error, will remove.

service-front/module/Application/src/Services/SiriusApiService.php
```diff
@@ -162,17 +163,18 @@ public function abandonCase(array $data, Request $request): array
...
+     * @param SiriusDocument $systemType
      * @param Request $request
-     * @param string $base64suffix
+     * @param string $pdfSuffixBase64 Optional. PDF file in base-64 format, if provided
+     *                                will be added to the generated letter.
      * @return array
      * @throws GuzzleException
      */
-    public function sendPdf(
+    public function sendDocumentRequest(
```
gregtyler, line 173: Nit, but I don't think `Request` adds much here?
```suggestion
    public function sendDocument(
```

PR #276, opened 2024-10-22
Feature/id 370 add =experian fraud check#major

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/service-front/module/Application/src/Helpers/FormProcessorHelper.php
+++ b/service-front/module/Application/src/Helpers/FormProcessorHelper.php
@@ -255,13 +255,13 @@ class FormProcessorHelper
             } elseif ($fraudCheck['decision'] === 'NODECISION') {
                 $template = $templates['thin_file'];
             } elseif ($fraudCheck['decision'] === 'STOP') {
-                $template = $templates['fraud'];
+                throw new OpgApiException('Stop response received from fraud check service.');
             } else {
                 $template = $templates['fail'];
             }
             $this->opgApiService->updateCaseSetDocumentComplete($uuid);
         } else {
-            $template = $templates['fail'];
+            throw new OpgApiException('Unknown response received from fraud check service.');
         }
         return $template;
     }
--- a/service-front/module/Application/src/Helpers/FormProcessorHelper.php
+++ b/service-front/module/Application/src/Helpers/FormProcessorHelper.php
@@ -37,6 +37,7 @@ class FormProcessorHelper
             $template = $validDln === 'PASS' ? $templates['success'] : $templates['fail'];
             if ($validDln === 'PASS') {
                 $this->opgApiService->updateCaseSetDocumentComplete($uuid);
+                $this->opgApiService->requestFraudCheck($uuid);
             }
         }
         return new FormProcessorResponseDto(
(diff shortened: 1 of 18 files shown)


service-front/module/Application/src/Helpers/FormProcessorHelper.php
```diff
@@ -242,4 +234,35 @@ public function processDataForm(array $params): string
...
+            if (
+                $fraudCheck['decision'] === 'ACCEPT' ||
+                $fraudCheck['decision'] === 'CONTINUE' ||
+                $fraudCheck['decision'] === 'REFER'
+            ) {
+                $template = $templates['success'];
+            } elseif ($fraudCheck['decision'] === 'NODECISION') {
+                $template = $templates['thin_file'];
+            } elseif ($fraudCheck['decision'] === 'STOP') {
+                $template = $templates['fraud'];
```
gregtyler, line 258: Per the ticket, this should throw an exception because we haven't decided what we're doing with STOP responses yet.

service-front/module/Application/src/Helpers/FormProcessorHelper.php
```diff
@@ -242,4 +234,35 @@ public function processDataForm(array $params): string
...
+            } else {
+                $template = $templates['fail'];
```
gregtyler, line 260: Similarly, if we receive an unknown response from Experian then we should throw an exception since it indicates a problem with the service (ours or theirs).

service-front/module/Application/src/Services/OpgApiService.php
```diff
@@ -366,4 +366,21 @@ public function getServiceAvailability(): DependencyCheck
...
+        $url = sprintf("/cases/%s/request-fraud-check", $uuid);
+
+        try {
+            $this->makeApiRequest($url, 'GET');
+        } catch (\Exception $exception) {
+            throw new OpgApiException($exception->getMessage());
+        }
+
+        if (empty($this->responseData)) {
+            throw new OpgApiException('Service availability data missing!');
```
gregtyler, line 381: Error message seems wrong here

service-front/module/Application/src/Helpers/FormProcessorHelper.php
```diff
@@ -34,10 +34,8 @@ public function processDrivingLicenceForm(
 
             $variables['dln_data'] = $formArray;
             $validDln = $this->opgApiService->checkDlnValidity($formArray['dln']);
-            $template = $validDln === 'PASS' ? $templates['success'] : $templates['fail'];
-            if ($validDln === 'PASS') {
-                $this->opgApiService->updateCaseSetDocumentComplete($uuid);
-            }
+
+            $template = $this->fraudCheck($validDln, $uuid, $templates);
```
gregtyler, line 38: I have a slight concern that doing this inside the individual ID provider checks creates too much coupling and makes the code for processing someone's ID more complicated than necessary.

Another option would be that a successful result in any of these takes you to `/:uuid/fraud-check` which does the API call then redirects them to KBVs or a thin file error, which would keep the code for identity checking and fraud checking separate. But that's more a suggestion than a dealbreaker.

spinhalf-com: I've pulled the fraud check back into the controller and it's all working ok, but it's increasingly revealing that the formProcessor, instead of being a useful way to abstract and reduce complexity, is mostly doing the opposite. I only introduced it to begin with because I didn't want fat and complicated controller methods, but it hasn't prevented that happening anyway.

sixdaysandy (approved):

PR #283, opened 2024-11-04
Id 359/send result to sirius

davidbridgwood:
## Purpose

Send the results from the KBV questions (pass/fail) to sirus with the `EventSender` and also save to `CaseData`

Fixes [ID-359](https://opgtransform.atlassian.net/browse/ID-359)

## Approach

`CaseOutcomeCalculator` (though i think this may need renaming), to update `CaseData` and send result to Sirius via `EventSender` - used by the KBV and PO routes.

## Learning

## Checklist

* [X] I have performed a self-review of my own code
* [X] I have added relevant logging with appropriate levels to my code
* [X] I have added tests to prove my work

[ID-359]: https://opgtransform.atlassian.net/browse/ID-359?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service-api/module/Application/src/Helpers/CaseOutcomeCalculator.php
```diff
@@ -0,0 +1,36 @@
...
+class CaseOutcomeCalculator
+{
+    public function __construct(
+        private readonly DataWriteHandler $dataHandler,
+        private readonly LoggerInterface $logger,
+        private readonly EventSender $eventSender,
+    ) {
+    }
+
+    public function updateSendIdentityCheck(CaseData $caseData, string $time): void
```
mohammedrmoj, line 21:
```suggestion
    /**
     * @param CaseData $caseData
     * @param string $time
     * @return void
     */
    public function updateSendIdentityCheck(CaseData $caseData, string $time): void
```

mohammedrmoj (approved):
Left one recommendation adding a DocBlock as I see these being added elsewhere. Looks good to me though.

service-api/module/Application/src/Helpers/CaseOutcomeCalculator.php
gregtyler: We generally only add docblocks to expand on types where PHP doesn't do specifics, e.g.:
```
/**
 * @return array{success: bool}
 */
public function saveAnswer(string $answer): array
```
Arrays and generics are the most common case, now that PHP's native typing is pretty well developed

mohammedrmoj: Understood

PR #285, opened 2024-11-05
Feature/id 393 switchable flows#minor

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


e2e-tests/cypress/e2e/experian/certificate-provider.cy.ts
```diff
@@ -0,0 +1,97 @@
+describe("Identify a Certificate Provider with a fraud checck", () => {
```
gregtyler, line 1: End to end tests are a very inefficient way of testing logic: these three CP tests all run through several of the same steps, and are ultimately just testing the outcome of `FormProcessorHelper->processTemplate()` which already has unit tests checking the same thing. The donor tests are similarly doing three more scenarios which test the same thing again.

There's benefit in having an end-to-end scenario that checks the fraud check integration is working, but that already happens in the general journey flows (e.g. "lets you identify with National Insurance number").

So I reckon both of these cypress files can be deleted since they're not testing anything that isn't already covered in other tests.

spinhalf-com: I'm not sure I wholly agree: the experian specific tests use the different LPAs that the new imposter mocks contain so that all the fraud check and KBV responses can be tested. I am happy though to move these tests out of the application so that they don't run in the pipeline, I'll just use them for local dev only. I do find them very useful for catching the occasional unexpected side effect.

gregtyler: Ultimately though, I think it's still testing the same thing as the unit test: if the remote responds with a "REFER", it should do X.

Moving it out makes sense to me. If they're regularly catching things that wouldn't be noticed otherwise, that suggests we're missing unit tests that should have identified it.

service-api/module/Application/src/Factories/SsmHandlerFactory.php
```diff
@@ -0,0 +1,27 @@
...
+     * @param null|array<mixed> $options
+     */
+    public function __invoke(ContainerInterface $container, $requestedName, ?array $options = null): SsmHandler
+    {
+        /** @var array{"aws": array<string, string|bool>} $config */
+        $config = $container->get('Config');
+
+        $ssmClient = new SsmClient($config['aws']);
+
+        return new SsmHandler($ssmClient);
```
gregtyler, line 25: There's already a factory for SsmClient, so you can just fetch it from the container:
```suggestion
        return new SsmHandler($container->get(SsmClient::class));
```

service-api/module/Application/src/Fixtures/SsmHandler.php
```diff
@@ -0,0 +1,27 @@
...
+    {
+    }
+
+    public function getParameter(string $paramName): array
+    {
+        $parameter = $this->ssmClient->getParameter([
+            'Name' => $paramName
+        ])->toArray();
+
+        if (is_string($parameter['Parameter']['Value'])) {
```
gregtyler, line 21: `$parameter['Parameter']['Value']` is always a string, even when it's not JSON, so this will cause an exception if you used it on a non-JSON parameter.

I'd suggest renaming the method to `getJsonParameter` to be clear about what it's doing.

service-api/module/Application/src/Fixtures/SsmHandler.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Fixtures;
```
gregtyler, line 5: This shouldn't be in the fixtures folder/namespace, but probably in `Application\Aws` instead?

service-api/module/Application/src/Controller/Factory/HealthcheckControllerFactory.php
```diff
@@ -31,13 +35,19 @@ public function __invoke(
...
+
         /**
          * @psalm-suppress PossiblyFalseArgument
          */
         return new HealthcheckController(
             $container->get(DataQueryHandler::class),
-            $container->get(SsmClient::class),
+            $container->get(SsmHandler::class),
             $ssmServiceAvailability,
+            $logger,
```
gregtyler, line 50: There's already a logger setup, so you can fetch it from the container rather than creating a new one:
```suggestion
            $container->get(LoggerInterface::class),
```

gregtyler (approved):

PR #289, opened 2024-11-06
Remove decisionText and responseHeader from ResponseDTO

gregtyler:
They're not used for anything elsewhere in the code.

#patch


mohammedrmoj (approved):
Looks good to me

PR #292, opened 2024-11-07
ID-398 - Fix Paper ID "Exit ID" not working

mohammedrmoj:
## Purpose

Exit ID link was not taking into account the base path: `/lpa/identity-check`

Fixes https://opgtransform.atlassian.net/browse/ID-398

## Approach

Using twig filter `basepath` which grabs it from the environment variable `PREFIX`

Tested locally and on an isolated deployment at https://id-398.development.sirius.opg.service.justice.gov.uk/


mohammedrmoj:
I tried to avoid having to any PHP for this one. One liner in twig.

davidbridgwood (approved):

spinhalf-com (approved):

PR #293, opened 2024-11-08
vouching route - confirm vouching

davidbridgwood:
## Purpose

Create the start of the vouching route with the "confirm vouching" page

Fixes [ID-265](https://opgtransform.atlassian.net/browse/ID-265)

## Approach

Setup a `VouchingFlowController` and allow for a new personType (voucher). When the process is started with `personType === voucher` then you are directed to the "confirm vouching" page.

Have added a `vouchingFor` field to `CaseData` which is populated with the name of the donor linked to the LPA. I also needed to make the fields associated with the donor in the other journeys (dob, name, address) non-required, defaulting to null/empty as this is info which will be provided by the voucher later in the vouching journey. This has had some knock on impacts on some validation checks and tests.

## Checklist

[ID-265]: https://opgtransform.atlassian.net/browse/ID-265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
+++ b/service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
@@ -48,7 +48,6 @@ class ConfigBuilder
         $saaConfig['LocationDetails'] = [
             'LocationIdentifier' => '1',
             'UKLocation' => [
-                // or de we want to trow an error if address is empty??
                 'HouseName' => $case->address["line1"] ?? '',
                 'Street' => $case->address["line2"] ?? '',
                 'District' => $case->address["line3"] ?? '',
--- a/service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
+++ b/service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
@@ -48,7 +48,8 @@ class ConfigBuilder
         $saaConfig['LocationDetails'] = [
             'LocationIdentifier' => '1',
             'UKLocation' => [
-                'HouseName' => $case->address["line1"],
+                // or de we want to trow an error if address is empty??
+                'HouseName' => $case->address["line1"] ?? '',
                 'Street' => $case->address["line2"] ?? '',
                 'District' => $case->address["line3"] ?? '',
                 'PostTown' => $case->address["town"] ?? '',
--- a/service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
+++ b/service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
@@ -45,6 +45,10 @@ class ConfigBuilder
         $saaConfig['ApplicationData'] = [
             'SearchConsent' => 'Y',
         ];
+
+        if (! isset($case->address)) {
+            throw new RuntimeException('Address has not been set');
+        }
         $saaConfig['LocationDetails'] = [
             'LocationIdentifier' => '1',
             'UKLocation' => [
--- a/service-front/module/Application/src/Controller/IndexController.php
+++ b/service-front/module/Application/src/Controller/IndexController.php
@@ -91,7 +91,7 @@ class IndexController extends AbstractActionController
      */
     private function processLpaResponse(string $type, array $data): array
     {
-        if (in_array($type, array('donor', 'voucher'))) {
+        if (in_array($type, ['donor', 'voucher'])) {
             if (! empty($data['opg.poas.lpastore'])) {
                 $address = (new AddressProcessorHelper())->processAddress(
                     $data['opg.poas.lpastore']['donor']['address'],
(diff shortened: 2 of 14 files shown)


service-front/module/Application/src/Controller/VouchingFlowController.php
```diff
@@ -0,0 +1,55 @@
...
+         * @psalm-suppress InvalidArrayOffset
+         */
+        $view->setVariable('vouching_for', $details_data["vouchingFor"]);
+        $view->setVariable('form', $form);
+
+        if ($this->getRequest()->isPost()) {
+            $formData = $this->getRequest()->getPost();
+
+            if (isset($formData['try_different'])) {
+                return $this->redirect()->toUrl(
```
davidbridgwood, line 43: should i be using `toRoute` instead of `toUrl` here - I did give it a go but couldn't make a success of it.

mohammedrmoj: So this doesn't work?
```
return $this->redirect()->toRoute("start", [
    'personType' => 'donor',
    'lpas' => $details_data['lpas']
]);
```

service-front/module/Application/src/Controller/IndexController.php
```diff
@@ -55,8 +55,12 @@ public function startAction(): Response
         $type = $this->params()->fromQuery("personType");
 
         if (! $this->lpaFormHelper->lpaIdentitiesMatch($lpas, $type)) {
-            $personTypeDescription = $type === 'donor' ? "Donors" : " Certificate Providers";
-            throw new HttpException(400, "These LPAs are for different " . $personTypeDescription);
+            $personTypeDescription = [
+                'donor' => "Donors",
+                'certificateProvidor' => "Certificate Providers",
```
mohammedrmoj, line 60: Is typo on purpose?
(certificateProvidor -> certificateProvider)

service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
```diff
@@ -48,7 +48,7 @@ public function buildSAARequest(CaseData $case): array
         $saaConfig['LocationDetails'] = [
             'LocationIdentifier' => '1',
             'UKLocation' => [
-                'HouseName' => $case->address["line1"],
+                'HouseName' => $case->address["line1"] ?? '',
```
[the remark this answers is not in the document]
mohammedrmoj: If we did then how would it be handled? Redirect to an address input screen?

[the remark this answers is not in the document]
davidbridgwood: i think any error thats thrown get handled by the framework so would redirect to a generic error page.

service-front/module/Application/src/Controller/VouchingFlowController.php
davidbridgwood: nope - and i couldn't figure out why,

service-front/module/Application/src/Controller/IndexController.php
davidbridgwood: good spot!

service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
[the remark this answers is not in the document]
mohammedrmoj: The problem with that is it doesn't give the user an opportunity to do something about it. What happens if we submit an empty address to experian? Does it accept it?

service-front/module/Application/src/Controller/VouchingFlowController.php
mohammedrmoj: Looking at this further... Wouldn't 
```
 implode("lpas[]=", $details_data['lpas'])
 ```
 Result in something more like
 ```
M-XYXY-YAGA-35G3lpas[]=M-XYXY-YAGA-35G2lpas[]=M-XYXY-YAGA-35G3
 ```
 ?
 
 I think you would need:
 ```
 implode("&lpas[]=", $details_data['lpas'])
 ```
 
 Note the extra `&`

Comment (mohammedrmoj) [reply service-front/module/Application/src/Controller/VouchingFlowController.php:43]: So 
```
"/start?personType=donor&lpas[]=" . implode("&lpas[]=", $details_data['lpas'])
```

service-api/module/Application/src/Experian/IIQ/ConfigBuilder.php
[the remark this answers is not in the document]
mohammedrmoj: I think your instincts are right. Given that we don't expect it to happen I would throw an exception. Maybe create a new one like MissingExperianArgumentException so we can catch it and handle if needed later on.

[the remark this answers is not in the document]
davidbridgwood: i've just added in a `RuntimeException` as that is consistent with whats already used in the same function.

PR #312, opened 2024-11-21
Bugfix/id 423 cp address lookup remote envs

mohammedrmoj:
## Purpose

https://opgtransform.atlassian.net/browse/ID-423

Fixes 

On remote environments:

If you enter a non-existent invalid postcode, you get a “there was a problem with this service” error page rather than a helpful error message

After selecting an address, you also get a “there was a problem with this service” error page

## Approach

Added postcode check (using postcode lookup) to enter postcode page to ensure that the postcode exists and display an error if it does not. It will not proceed to the next screen if the postcode does not return any results or the response is a 400 error.

## Checklist

### DIFF
--- a/service-front/module/Application/view/application/pages/cp/select_address.twig
+++ b/service-front/module/Application/view/application/pages/cp/select_address.twig
@@ -10,7 +10,7 @@
             What is the address on the ID document?
         {% endblock %}
     </h1>
-    <form method="POST" action="./select-address">
+    <form method="POST">
         <div class="govuk-form-group {% if form.get('address_json').messages %}govuk-form-group--error{% endif %}">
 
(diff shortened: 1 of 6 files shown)


service-front/module/Application/view/application/pages/cp/select_address.twig
```diff
@@ -10,7 +10,7 @@
             What is the address on the ID document?
         {% endblock %}
     </h1>
-    <form method="POST" action="./select-address">
```
mohammedrmoj, line 13: This fixes the form submitting to `select-address/select-address` on remote.. It still works locally too.

spinhalf-com (approved):

PR #319, opened 2024-11-25
ID-378 voucher address pages

davidbridgwood:
## Purpose

create the pages needed to enter the vouchers address details and raise en error if it matches that of the donor.

Fixes [ID-378](https://opgtransform.atlassian.net/browse/ID-378)

## Checklist

[ID-378]: https://opgtransform.atlassian.net/browse/ID-378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service-api/module/Application/config/module.config.php
```diff
@@ -320,6 +320,16 @@
...
+            'save_address_to_case' => [
+                'type' => Segment::class,
+                'options' => [
+                    'route' => '/cases/:uuid/save-address-to-case',
+                    'defaults' => [
+                        'controller' => Controller\IdentityController::class,
+                        'action' => 'saveAddressToCase',
+                    ],
+                ],
+            ],
```
mohammedrmoj, line 332: I was just wondering regarding choosing the method and route, is there any reason we're not trying to be more restful 

So for example a restful endpoint for this would be: 
```
POST /cases/:uuid/addresses
```

Not singling this particular route out as I can see we're not following a restful approach in other places as well.

service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -320,6 +320,42 @@ public function removeCaseLpaAction(): JsonModel
...
+
+            return new JsonModel($response);
+        }
+
+        try {
+            $this->dataHandler->updateCaseData(
+                $uuid,
+                'address',
+                $data,
+            );
```
mohammedrmoj, line 345: This will need to be visited as soon as claimedIdentity has been merged.

mohammedrmoj (approved):
Might be worth check sonar cloud for some issues raised there. 

We'll need to make sure we revisit updateCaseData for address since its not being added to the `claimedIdentity` now, unless I missed something?

service-api/module/Application/config/module.config.php
davidbridgwood: its a good question - i'm not really sure, i guess i've just been following the pattern already in place?

mohammedrmoj (dismissed):

mohammedrmoj (approved):

PR #320, opened 2024-11-26
Feature/id 417 create claimed identity object

mohammedrmoj:
## Purpose

Currently information about the identity the caller is claiming to be is stored at the top-level of the CaseData object. To improve the structure of our data-model we have agreed to create a sub-object to store the information about the claimed-identity in one place.

Move the following fields from CaseData into an object of there own and update all impacted code.

- firstName
- lastName
- dob
- address

(Will be adding and replacing alternateAddress with professionalAddress in another ticket)

Fixes ID-417

## Approach

Moved the data on the API and refactored relevant code and unit tests

## Checklist


service-front/module/Application/src/Services/OpgApiService.php
```diff
@@ -72,6 +72,14 @@ public function getDetailsData(string $uuid): array
     {
         try {
             $response = $this->makeApiRequest('/identity/details?uuid=' . $uuid);
+
```
davidbridgwood, line 75: Is the intention to do the front-end refactor as part of another ticket? (makes sense). Has it been created?

davidbridgwood (approved):

PR #351, opened 2024-12-19
Feature/id 380/add donors/#minor

davidbridgwood:
## Purpose

Add the "confirm-donors" and "add-donor" pages to the vouching route. This includes logic to compare the name/dob/address of the person vouching with actors on the LPAs being added to the ID check.
You can test this locally as described in [this comment](https://opgtransform.atlassian.net/browse/ID-379?focusedCommentId=97015).

Fixes [ID-380](https://opgtransform.atlassian.net/browse/ID-380) and [ID-379](https://opgtransform.atlassian.net/browse/ID-379)

## Approach

[service-front/module/Application/src/Helpers/AddDonorFormHelper.php](https://github.com/ministryofjustice/opg-paper-identity/pull/351/files#diff-fed328822b24d095239182503ec1eddb366a93552ae2f3326bd0a6ad7ddf68ec) holds most of the logic to check the LPAs retuned.

## Checklist

[ID-380]: https://opgtransform.atlassian.net/browse/ID-380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ID-379]: https://opgtransform.atlassian.net/browse/ID-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service-front/module/Application/src/Controller/VouchingFlowController.php
```diff
@@ -385,6 +399,91 @@ public function enterAddressManualAction(): ViewModel|Response
...
+            $lpaDetails[$lpa] = [
+                'name' => $donorName,
+                'type' => $type,
+            ];
+        }
+
+        $view = new ViewModel();
+
+        $view->setVariable('lpas', $detailsData['lpas']);
+        $view->setVariable('lpa_count', count($detailsData['lpas']));
```
mohammedrmoj, line 426: If we're passing the entire $detailsData into the view then do we need to pass all the individual values inside it too?

service-front/module/Application/src/Helpers/AddDonorFormHelper.php
```diff
@@ -0,0 +1,332 @@
...
+            $response["problem"] = true;
+            $response["message"] = "No LPA Found.";
+        }
+        return $response;
+    }
+
+    /**
+     * @return LpaIdMatchCheck
+     */
+    public function checkLpaIdMatch(array $lpa, array $detailsData): array
```
mohammedrmoj, line 128: Add @params with types for (array $lpa, array $detailsData) to docBlock?

service-front/module/Application/src/Helpers/AddDonorFormHelper.php
```diff
@@ -0,0 +1,332 @@
...
+                ]
+            ];
+        }
+        return $response;
+    }
+
+    /**
+     * @return CheckedLpa[]
+     */
+    private function checkLpas(array $lpasData, array $detailsData): array
```
mohammedrmoj, line 219: Add @params with types for (array $lpa, array $detailsData) to docBlock?

service-front/module/Application/src/Helpers/AddDonorFormHelper.php
```diff
@@ -0,0 +1,332 @@
...
+
+            $lpas[] = $result;
+        }
+        return $lpas;
+    }
+
+    /**
+     * @return ProcessedLpas
+     */
+    public function processLpas(array $lpasData, array $detailsData): array
```
mohammedrmoj, line 243: Add @params with types for (array $lpasData, array $detailsData) to docBlock?

service-front/module/Application/src/Helpers/AddDonorFormHelper.php
```diff
@@ -0,0 +1,332 @@
...
+        if (empty($lpasData)) {
+            $response["problem"] = true;
+            $response["message"] = "This LPA has already been added to this identity check.";
+            return $response;
+        }
+
+        $lpas = $this->checkLpas($lpasData, $detailsData);
+
+        // if there is 1 LPA returned and there is a problem then we just flag the problem
+
```
mohammedrmoj, line 272:
```suggestion
        // if there is 1 LPA returned and there is a problem then we just flag the problem
```

service-front/module/Application/src/Helpers/VoucherMatchLpaActorHelper.php
```diff
@@ -92,9 +93,9 @@ private function compareDob(string $dob, array $actor): bool
      * @param string $firstName
      * @param string $lastName
      * @param string $dob
-     * @return Actor[]
+     * @return false|Actor
```
mohammedrmoj, line 96:
```suggestion
     * @return false|Actor[]
```
Seems it should still be an array here

mohammedrmoj (approved):
Have approved, but have added some recommendations.

service-front/module/Application/src/Helpers/VoucherMatchLpaActorHelper.php
davidbridgwood: It actually is just a single `Actor` rather than an array of them returned now. I changed this as we only ever surface the first match to users (and its only very rare circumstances where there would be >1 anyways) so its simpler downstream to just return one here.

mohammedrmoj (approved):

PR #374, opened 2025-01-16
Feature/id 375 dwp client citizen data#minor

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-api/module/Application/src/DWP/DwpApi/DTO/CitizenRequestDTO.php
```diff
@@ -0,0 +1,107 @@
...
+             */
+            if ($this->caseData->idMethodIncludingNation->id_method !== 'NATIONAL_INSURANCE_NUMBER') {
+                throw new DwpApiException('Identity method is not a national insurance number');
+            }
+
+            /**
+             * @psalm-suppress PossiblyNullArrayAccess
+             * @psalm-suppress PossiblyNullPropertyFetch
+             * @psalm-suppress PossiblyNullPropertyAssignmentValue
+             */
```
davidbridgwood, line 34: it might be neater to put these in the function docblock rather than repeating them for each variable.
Though I wonder if it would be better to throw an exception if any of them are null as they shouldn't be at this point in the journey.

davidbridgwood (commented):
@spinhalf-com - have added a couple of small comments, might also be worth a quick review from someone more familiar with the auth/dwp api stuff to check i haven't missed anything.

What is the behaviour meant to be when the api returns that there is no-match or a nino is invalid? When testing locally I get stuck if i enter anything other than "NP 11 22 33 C" but don't get any sort of error/indication that something is not right.

Would be good to [update the README too](https://github.com/ministryofjustice/opg-paper-identity/?tab=readme-ov-file#passport-driving-license-and-national-insurance-number-lookup) as the behaviour of the nino mock is changing.

service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -154,19 +158,36 @@ public function detailsAction(): JsonModel
...
         $data = json_decode($this->getRequest()->getContent(), true);
+        $caseData = $this->dataQueryHandler->getCaseByUUID($uuid);
+        /**
+         * @psalm-suppress PossiblyNullPropertyFetch
+         */
+        $idMethodValues = $caseData->idMethodIncludingNation;
+        /**
+         * @psalm-suppress PossiblyNullPropertyAssignment
+         */
+        $idMethodValues->id_value = $data['nino'];
```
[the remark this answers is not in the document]
spinhalf-com: We need to save it now that we're making it part of a request to the 3rd party api.

service-api/module/Application/config/module.config.php
```diff
@@ -502,7 +506,7 @@
                 $serviceLocator->get(LoggerInterface::class)
             ),
             LoggerInterface::class => LoggerFactory::class,
-            NinoValidatorInterface::class => NinoValidatorFactory::class,
+//            NinoValidatorInterface::class => NinoValidatorFactory::class,
```
mohammedrmoj, line 509: Remove this?

service-api/module/Application/src/DWP/AuthApi/DTO/ResponseDTO.php
```diff
@@ -0,0 +1,43 @@
...
+
+    public function accessToken(): string
+    {
+        return $this->accessToken;
+    }
+
+    public function expiresIn(): string
+    {
+        return $this->expiresIn;
+    }
```
gregtyler, line 24: Similarly, you could make the properties `public readonly` rather than having a passthrough getter

service-api/module/Application/src/DWP/DwpApi/DTO/CitizenResponseDTO.php
```diff
@@ -0,0 +1,66 @@
...
+
+    public function version(): string
+    {
+        return $this->version;
+    }
+
+    public function raw(): array
+    {
+        return $this->raw;
+    }
```
gregtyler, line 56: I don't think we're using the raw response so I don't think we should store/expose it, since it seems like an easy way to accidentally log personal information that we don't want

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,214 @@
...
+                'Bearer %s',
+                $this->authApiService->retrieveCachedTokenResponse()
+            ),
+            'Correlation-Id' => '',
+            'Policy-Id' => '',
+            'Instigating-User-Id' => ''
+        ];
+    }
+
+    public function validateNino(CaseData $caseData): array
```
gregtyler, line 56: Rather than returning an unstructured array, could this just return a boolean of whether it was matched or not?

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,214 @@
...
+    public function compareRecords(
+        CaseData $caseData,
+        DetailsResponseDTO $detailsResponseDTO,
+        CitizenResponseDTO $citizenResponseDTO
+    ): array {
+        if ($caseData->idMethodIncludingNation) {
+            if (
+                $citizenResponseDTO->matchScenario() !== 'Matched on NINO' ||
+                $detailsResponseDTO->verified() !== 'verified'
+            ) {
```
gregtyler, line 81: I think we need to check here that the full nino matches `$detailsResponseDTO->nino`?

service-api/module/Application/test/ApplicationTest/Services/DWP/DwpApi/DwpApiServiceTest.php
```diff
@@ -0,0 +1,569 @@
...
+                "line3" => "",
+                "town" => "Southamption",
+                "line2" => "",
+                "line1" => "18 BOURNE COURT"
+            ],
+            "professionalAddress" => null
+        ]
+    ];
+
+    private const DETAILS_RESPONSE = [
```
gregtyler, line 73: I don't think there's any point including any of the data we don't use in this: it makes the test hard to read and makes it look like we're depending on fields which we're not.

service-api/module/Application/src/Controller/IdentityController.php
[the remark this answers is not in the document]
gregtyler: Can we just pass the NINO as an extra argument to `validateNino` (i.e. `$this->dwpApiService->validateNino($caseData, $data['nino']);`? I'd rather avoid storing it any longer than necessary unless we have a specific need.

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,248 @@
...
+                new CitizenRequestDTO($caseData, $nino)
+            );
+            $detailsResponseDTO = $this->makeCitizenDetailsRequest(
+                new DetailsRequestDTO($citizenResponseDTO->id()),
+                $nino
+            );
+
+            return $this->compareRecords($caseData, $detailsResponseDTO, $citizenResponseDTO, $nino);
+        } catch (\Exception $exception) {
+            $this->logger->error('DwpApiException: ' . $exception->getMessage());
```
spinhalf-com, line 72: Add in logging of correlation ID here

docker/dwp/openapi.yml
```diff
@@ -149,7 +143,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorData'
-  /v2/citizens/match/{guid}:
+  /capi/v2/citizens/{guid}:
```
spinhalf-com, line 146: Redact in accordance with DWP contract

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,248 @@
...
+     * @throws DwpApiException
+     */
+    public function makeCitizenMatchRequest(
+        CitizenRequestDTO $citizenRequestDTO,
+    ): CitizenResponseDTO {
+        $this->authCount++;
+        $responseArray = [];
+        try {
+            $postBody = $this->constructCitizenRequestBody($citizenRequestDTO);
+            $this->logger->info('MATCH_POSTBODY: ', $postBody);
```
mohammedrmoj, line 129: I think we'd want to remove this

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,248 @@
...
+            $response = $this->guzzleClient->request(
+                'POST',
+                $this->matchPath,
+                [
+                    'headers' => $this->makeHeaders(),
+                    'json' => $postBody
+                ]
+            );
+            $responseArray = json_decode($response->getBody()->getContents(), true);
+            $this->logger->info("DWP_MATCH_RESPONSE: " . json_encode($responseArray, JSON_THROW_ON_ERROR));
```
mohammedrmoj, line 140: Also needs clean up?

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,248 @@
...
+
+    public function makeNinoFragment(string $nino): string
+    {
+        $nino = str_replace(" ", "", $nino);
+        return substr($nino, (strlen($nino) - 5), - 1);
+    }
+
+    public function makeFormattedPostcode(string $postcode): string
+    {
+        $cleanPostcode = preg_replace("/[^A-Za-z0-9]/", '', $postcode);
```
mohammedrmoj, line 197: To avoid the need for the suppress try: 
```
$cleanPostcode = (string) preg_replace("/[^A-Za-z0-9]/", '', $postcode);
```

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,248 @@
...
+            );
+            $responseArray = json_decode($response->getBody()->getContents(), true);
+            $this->logger->info("DWP_MATCH_RESPONSE: " . json_encode($responseArray, JSON_THROW_ON_ERROR));
+        } catch (ClientException $clientException) {
+            if (
+                $clientException->getResponse()->getStatusCode() == Response::STATUS_CODE_401 &&
+                $this->authCount < 2
+            ) {
+                $this->authApiService->authenticate();
+                $this->makeCitizenMatchRequest($citizenRequestDTO);
```
mohammedrmoj, line 147: Don't we need to return the value here?
```
return $this->makeCitizenMatchRequest($citizenRequestDTO);
```

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,248 @@
...
+            );
+            $responseArray = json_decode($response->getBody()->getContents(), true);
+            $this->logger->info("DWP_DETAILS_RESPONSE: " . json_encode($responseArray, JSON_THROW_ON_ERROR));
+        } catch (ClientException $clientException) {
+            if (
+                $clientException->getResponse()->getStatusCode() == Response::STATUS_CODE_401 &&
+                $this->authCount < 2
+            ) {
+                $this->authApiService->authenticate();
+                $this->makeCitizenDetailsRequest($detailsRequestDTO, $nino);
```
mohammedrmoj, line 232: The value needs to returned here.

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
spinhalf-com: No, it's in the responseDto at the end of the function

gregtyler: I think you still need to set `$responseArray`, otherwise the value is never captured from the second request
```suggestion
                $responseArray = $this->makeCitizenMatchRequest($citizenRequestDTO);
```

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,243 @@
...
+            }
+        } catch (\Exception $exception) {
+            throw new DwpApiException($exception->getMessage());
+        }
+        return new CitizenResponseDTO(
+            $responseArray
+        );
+    }
+
+    public function constructCitizenRequestBody(
```
gregtyler, line 160: CitizenRequestDTO exists to translate our local data to something acceptable for the citizen request endpoint, so I suspect this function should be inside the DTO class

service-api/module/Application/src/Model/Entity/IdMethodIncludingNation.php
```diff
@@ -32,6 +32,9 @@ class IdMethodIncludingNation extends Entity
     #[Annotation\Validator(NotEmpty::class, options: [NotEmpty::NULL])]
     public string $id_country;
 
+    #[Annotation\Required(false)]
+    #[Annotation\Validator(NotEmpty::class, options: [NotEmpty::NULL])]
+    public ?string $id_correlation;
```
gregtyler, line 37: This is specific to DWP, so I feel the name should reflect that. It could even go inside its own subobject on CaseData (like CounterService and IdentityIQ)

service-api/module/Application/test/ApplicationTest/Services/DWP/DwpApi/DwpApiServiceTest.php
gregtyler, line 1: This test doesn't seem to check `validateNino` at any point, but that's arguably the most important thing it should check because that's the method it exposes to other services.

docker/dwp/dwp-config.yaml
```diff
@@ -17,18 +17,47 @@ resources:
...
+      statusCode: 422
+      file: data/ambiguousResponse.json
+
+  - path: "/capi/v2/citizens/match"
+    method: post
+    response:
+      statusCode: 404
+      file: data/notFoundResponse.json
+
+  - path: "/capi/v2/citizens/be62ed49-5407-4023-844c-97159ec80411"
```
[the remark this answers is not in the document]
spinhalf-com: It matches the foundResponse,json file

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,200 @@
...
+            $uri = $this->detailsPath . $detailsRequestDTO->id();
+            $response = $this->guzzleClient->request(
+                'GET',
+                $uri,
+                [
+                    'headers' => $this->makeHeaders(['Access-Level => 4']),
+                ]
+            );
+            $responseArray = json_decode($response->getBody()->getContents(), true);
+            $this->logger->info("DWP_DETAILS_RESPONSE: " . json_encode($responseArray, JSON_THROW_ON_ERROR));
```
mattmachell, line 177: Potential for PII leakage here.

service-api/module/Application/src/DWP/DwpApi/DwpApiService.php
```diff
@@ -0,0 +1,200 @@
...
+            if (
+                $clientException->getResponse()->getStatusCode() == Response::STATUS_CODE_401 &&
+                $this->authCount < 2
+            ) {
+                $this->authApiService->authenticate();
+                $this->makeCitizenDetailsRequest($detailsRequestDTO, $nino);
+            } else {
+                $response = $clientException->getResponse();
+                $responseBodyAsString = $response->getBody()->getContents();
+                $this->logger->error('GuzzleDwpApiException: ' . $responseBodyAsString);
```
mattmachell, line 188: Is the body always an error string? Depending on what is in the error contents, might also be potential for PII leakage here too.

service-api/docker-compose.env
```diff
@@ -25,3 +25,4 @@ YOTI_SESSION_DEADLINE=30
 SECRETS_MANAGER_PREFIX=local/paper-identity/
 EXPERIAN_CROSSCORE_BASE_URL=http://experian-crosscore-mock:8080/decisionanalytics/crosscore/cc_mock_route/
 EXPERIAN_CROSSCORE_AUTH_URL=http://experian-crosscore-auth-mock:8080/oauth2/experianone/v1/token
+DWP_SUPPRESS_CERTIFICATE=true
```
mattmachell, line 28: Might be worth adding a comment to indicate what this env var does in dev.

docker/sirius/data/donorDwp.json
```diff
@@ -0,0 +1,128 @@
...
+    "witnessedByCertificateProviderAt": "1942-01-14T13:22:49.0Z",
+    "witnessedByIndependentWitnessAt": "1930-09-17T01:21:25.0Z"
+  },
+  "opg.poas.sirius": {
+    "caseSubtype": "personal-welfare",
+    "donor": {
+      "addressLine1": "25 Queens Drive",
+      "addressLine3": "",
+      "dob": "05/09/1982",
+      "firstname": "Shaun",
```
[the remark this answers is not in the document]
gregtyler: Documentation still needs updated here

docker/dwp/dwp-config.yaml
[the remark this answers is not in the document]
gregtyler: Ah cheers, missed that that was already in the repo

service-api/module/Application/test/ApplicationTest/Services/DWP/DwpApi/DwpApiServiceTest.php
```diff
@@ -0,0 +1,461 @@
...
+            ]
+        ];
+        $successMockResponseDetailsData = static::DETAILS_RESPONSE;
+
+        $successMock = new MockHandler([
+            new GuzzleResponse(200, [], json_encode($successMockResponseMatchData, JSON_THROW_ON_ERROR)),
+            new GuzzleResponse(200, [], json_encode($successMockResponseDetailsData, JSON_THROW_ON_ERROR)),
+        ]);
+        $handlerStack = HandlerStack::create($successMock);
+        $successClient = new Client(['handler' => $handlerStack]);
```
gregtyler, line 437: Reflecting on this, this test would probably be better if it used a PHPUnit mock rather than a mock handler so that we could check that the correct endpoints are called. At the moment, it just tests that two HTTP requests are made but we don't check the path, headers or body being used.

gregtyler (approved):

PR #385, opened 2025-01-24
Handle DOB well

mohammedrmoj:
## Purpose

This PR adds validation and error handling for date of birth (DOB) input in the CP and vouch journey, ensuring users enter valid dates: if the DOB makes the person over 100 years old, a warning allows continuation but prompts confirmation; if the date is impossible (e.g., 29/02/1973) or in the future, an error message prevents continuation. Input fields are restricted to 2 characters for day/month and 4 for year, and an empty year field triggers the error "Enter a valid date" instead of referencing age. This aligns with existing logic and improves user guidance.

Fixes https://opgtransform.atlassian.net/browse/ID-441

## Approach

Updated controllers and Validation classes

## Checklist


davidbridgwood:
Hey @mohammedrmoj 
Have had a quick look at this and noticed a couple of high-level things.

I think the formatting when the dob is >100yrs needs to be slightly different (blue warning rather than red box and text), see the design [here](https://www.figma.com/design/pEYIyMBoWCGAvTclJFMOsq/Paper-ID-ideas?node-id=11706-33060&t=OW18jkRQULU4fWFA-4).

I've also noticed that if you input a day and month but leave the year blank you get a "Date of birth must be in the past." error but i think "Date of birth must be a valid date" would be more appropriate.

Is there anyway we can abstract the logic for the >100yrs warning/accept/continue so we don't just copy it in two places? And either way its probably worth adding a new unit-test to cover that scenario.

mohammedrmoj:
@davidbridgwood Do we have any examples of where warnings are displayed elsewhere on first submission?

davidbridgwood (dismissed):

PR #400, opened 2025-01-31
Prevent value being lost from Passport in date on form submission

mohammedrmoj:
## Purpose

To recreate - start an ID check, and on that page select the passport option and open the reveal and enter a date and click check. The page refreshes and the passport option is unchecked. 

This is frustrating because each user after they had checked would scroll down to the continue button and then get the error and then have to select the radio button again. 

Can this radio button stay as checked after the passport date has been checked please. 

Fixes https://opgtransform.atlassian.net/browse/ID-467

## Approach

`<input class="govuk-radios__input" id="inDateFalse" name="inDate" type="radio"
                                value="no" {{ (form.get('inDate').value == 'no') ? 'checked' : '' }}>`

### DIFF
--- a/e2e-tests/cypress/e2e/certificate_provider/certificate-provider-journey.cy.ts
+++ b/e2e-tests/cypress/e2e/certificate_provider/certificate-provider-journey.cy.ts
@@ -144,7 +144,7 @@ describe("Identify a Certificate Provider", () => {
         cy.contains("The passport needs to be no more than 18 months out of date");
 
         cy.get("#passport").clear();
-        cy.contains("Yes").click();
+        cy.get(".govuk-radios__label").contains("Yes").click();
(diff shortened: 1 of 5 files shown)


davidbridgwood:
@mohammedrmoj - from the ticket I think this also needs to be applied to the "[how will you confirm your identity](https://github.com/ministryofjustice/opg-paper-identity/blob/main/service-front/module/Application/view/application/pages/how_will_you_confirm.twig)" and "[which document to take to the post office](https://github.com/ministryofjustice/opg-paper-identity/blob/main/service-front/module/Application/view/application/pages/post_office/post_office_documents.twig)" pages.

davidbridgwood (approved):

davidbridgwood (approved):

PR #420, opened 2025-02-20
Added Court of protection route screen

mohammedrmoj:
https://opgtransform.atlassian.net/browse/ID-411


davidbridgwood (commented):
All looks good to me! I wonder if its worth adding a cypress test for this new route though?

davidbridgwood (dismissed):

davidbridgwood (approved):

PR #423, opened 2025-02-21
Feature/id 510 completed lpa error start page#minor

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-front/module/Application/src/Controller/IndexController.php
```diff
@@ -82,24 +91,37 @@ public function startAction(): Response
...
+                return $lpaStatusCheck->getStatus() === 'registered' ?
+                    "The identity check has already been completed" :
                     "ID check has status: " . $lpaStatusCheck->getStatus() . " and cannot be started";
-
-                throw new HttpException(400, $errorMessage);
             }
+            return 'ok';
         } catch (\Exception $exception) {
             throw new HttpException(400, $exception->getMessage());
         }
```
gregtyler, line 127: This method is a blending a bit too much together (whether it's startable, and what the error message should be) and it's only called in one place. I'd suggesting splitting it into two, but then you'd need to generate `$lpaStatusCheck` twice (once to return `->isStartable()`, once to calculate the message). Could the logic just be merged into `startAction`?

service-front/module/Application/src/Controller/DocumentCheckController.php
```diff
@@ -43,6 +43,14 @@ public function nationalInsuranceNumberAction(): ViewModel
         $form = $this->createForm(NationalInsuranceNumber::class);
         $detailsData = $this->opgApiService->getDetailsData($uuid);
 
+        /**
+         * @psalm-suppress InvalidArrayOffset
+         */
```
gregtyler, line 48: To avoid needing this suppressions, add `identityCheckPassed` to [the shape definition for CaseData](https://github.com/ministryofjustice/opg-paper-identity/blob/main/service-front/module/Application/src/Contracts/OpgApiServiceInterface.php#L24-L64)

service-front/module/Application/view/application/pages/cannot_start.twig
```diff
@@ -0,0 +1,18 @@
...
+            <svg class="moj-banner__icon" fill="currentColor" role="presentation" focusable="false"
+                 xmlns="http://www.w3.org/2000/svg" viewBox="0 0 25 25" height="25" width="25">
+                <path d="M13.7,18.5h-2.4v-2.4h2.4V18.5z M12.5,13.7c-0.7,0-1.2-0.5-1.2-1.2V7.7c0-0.7,0.5-1.2,1.2-1.2s1.2,0.5,1.2,1.2v4.8C13.7,13.2,13.2,13.7,12.5,13.7z M12.5,0.5c-6.6,0-12,5.4-12,12s5.4,12,12,12s12-5.4,12-12S19.1,0.5,12.5,0.5z"/>
+            </svg>
+            <div class="moj-banner__message">
+                <h4 class="govuk-heading-s govuk-!-margin-bottom-0">{{ message }}</h4>
+            </div>
+        </div>
+    </section>
+    <a href="{{ sirius_url }}" class="govuk-button">Close and return to Sirius</a>
```
gregtyler, line 17: I don't think `sirius_url` is set when you're rendering this in e.g. `nationalInsuranceNumberAction`?

To avoid having to juggle variables in the various controllers, you could create a new URL for this page and do a redirect instead? (i.e. DocumentCheckController would do `$this->redirect()->toRoute('root/cannot_start', ['uuid' => $uuid]);` rather than embedding the template).

spinhalf-com: Problem here is that there is no uuid created because the case creation is blocked. However the page design still requires the Return to Sirius link and still wants the name and personType displayed, so there has to be a separate method for getting that data that doesn't require a case.

gregtyler (approved):

PR #425, opened 2025-02-24
add code-coverage to unit-tests

davidbridgwood:
## Purpose

adds code-coverage so we can see areas where our testing could be improved - we are not setting a required threshold/failing a pipeline based on coverage at this point.

Fixes ###-####

## Approach

## Learning

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,7 +88,6 @@ jobs:
           directory: ./service-${{ matrix.service }}/build
           fail_ci_if_error: false
           flags: paper-id-${{ matrix.service }}
-          verbose: true
           token: ${{ secrets.CODECOV_TOKEN }}
   cypress:
     name: Cypress Tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,31 +80,12 @@ jobs:
           files: service-${{ matrix.service }}/build/*.xml
           test_file_prefix: +service-${{ matrix.service }}/
           check_name: ${{ matrix.service }} ${{ matrix.suite }} tests
-      - name: archive test results
-        uses: actions/upload-artifact@v4.6.0
-        with:
-          name: service-${{ matrix.service }}
-          path: service-${{ matrix.service }}/build
-        if: matrix.suite == 'unit-test'
-  code_coverage:
-    strategy:
-      matrix:
-        service: [api, front]
-    runs-on: ubuntu-latest
-    needs: test
-    steps:
-      - uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2 # pin@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -77,6 +80,35 @@ jobs:
...
+          path: service-${{ matrix.service }}/build
+        if: matrix.suite == 'unit-test'
+  code_coverage:
+    strategy:
+      matrix:
+        service: [api, front]
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      - uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2 # pin@v3
```
gregtyler, line 96: Better to use named versions so Renovate can auto-update
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/test.yml
```diff
@@ -77,6 +80,35 @@ jobs:
           files: service-${{ matrix.service }}/build/*.xml
           test_file_prefix: +service-${{ matrix.service }}/
           check_name: ${{ matrix.service }} ${{ matrix.suite }} tests
+      - name: archive test results
+        uses: actions/upload-artifact@v4.6.0
+        with:
+          name: service-${{ matrix.service }}
+          path: service-${{ matrix.service }}/build
+        if: matrix.suite == 'unit-test'
```
gregtyler, line 88: This may be something you tried and there's a reason it doesn't work, but I wonder if you could do the CodeCov upload here to avoid needing a save/download sequence.
```suggestion
      - name: service ${{ matrix.service }} codecov upload
        uses: codecov/codecov-action@v5
        if: matrix.suite == 'unit-test'
        with:
          name: paper-id-${{ matrix.service }}
          directory: ./service-${{ matrix.service }}
          fail_ci_if_error: false
          flags: paper-id-${{ matrix.service }}
          verbose: true
          token: ${{ secrets.CODECOV_TOKEN }}
```

.github/workflows/test.yml
```diff
@@ -77,6 +80,35 @@ jobs:
...
+          name: service-${{ matrix.service }}
+          path: service-${{ matrix.service }}
+      - name: service ${{ matrix.service }} codecov upload
+        uses: codecov/codecov-action@v5
+        with:
+          name: paper-id-${{ matrix.service }}
+          directory: ./service-${{ matrix.service }}
+          fail_ci_if_error: false
+          flags: paper-id-${{ matrix.service }}
+          verbose: true
```
gregtyler, line 110: Probably don't need verbosity now it's working?

.github/workflows/test.yml
davidbridgwood: ah this is nicer, have updated and seems to be working fine

gregtyler (approved):

PR #447, opened 2025-03-07
Feature/id 499 lpa helper function#minor

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


service-front/module/Application/src/Controller/PostOfficeFlowController.php
```diff
@@ -16,6 +16,7 @@
 use Application\Forms\PostOfficeSelect;
 use Application\Forms\PostOfficeSearch;
 use Application\Helpers\FormProcessorHelper;
+use Application\Helpers\SiriusDataProcessorHelper;
```
davidbridgwood, line 19: looks like you've missed making use of the new function in this file, could replace lines 266-275

service-front/module/Application/src/Helpers/SiriusDataProcessorHelper.php
```diff
@@ -135,4 +140,32 @@ public function processLpaResponse(string $type, array $data): array
 
         throw new HttpException(400, 'Person type "' . $type . '" is not valid');
     }
+
+    public function createLpaDetailsArray(
+        array $detailsData,
+        Request|RequestInterface $request,
+        bool $sirius = false
```
davidbridgwood, line 147: my understanding is that we should __always__ use the lpastore if it is present and only revert to sirius if we have to. In which case we don't need this argument and if there is anywhere in the code that we are not doing that already we can switch to that as part of this ticket.

spinhalf-com: The Vouching route appears to need this flag explicitly set in order to use the sirius section of the LPA.

davidbridgwood: yep - so i'm pretty sure that the vouching route is incorrectly deferring to the Sirius section of the LPA when it should be using the lpastore if its present. One of the reasons I thought this ticket would be useful was to iron out these inconsistencies.
@gregtyler - just pinging you to make sure my understanding is correct.

gregtyler: Yep, LPA Store data—if present—will always have more up-to-date data.

davidbridgwood (approved):

PR #449, opened 2025-03-10
Feature/id 391/cleanup data

davidbridgwood:
## Purpose

sets a TTL of 30 days on our dynamodb table to expire cases.

Fixes [ID-391](https://opgtransform.atlassian.net/browse/ID-391)

Required infra changes already made in https://github.com/ministryofjustice/opg-sirius-infrastructure/pull/2570

see [this thread](https://mojdt.slack.com/archives/C05SW1X69UH/p1741687761518269) for confirmation of which scenarios should result in a TTL being set.

## Approach

- add a `setTTL` and `unsetTTL` method to the `DataWriteHandler`
- `setTTL` is called at all terminal events.
- `unsetTTL` called when yoti-session is created so in the situation that someone has failed one method, which would trigger `setTTL` this is then unset)
- also cleaned up `IdentityController` to create single method to send events to sirius rather than a near identical method copied multiple times.

## Checklist

[ID-391]: https://opgtransform.atlassian.net/browse/ID-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


service-api/module/Application/src/Controller/IdentityController.php
```diff
@@ -813,12 +781,15 @@ public function sendVouchStartedAction(): JsonModel
...
-            "time" => $this->clock->now()->format('c'),
+            "reference" => "opg:" . $caseData->id,
             "actorType" => $caseData->personType,
             "lpaUids" => $caseData->lpas,
-            "state" => UpdateStatus::VouchStarted->value,
+            "time" => $this->clock->now()->format('c'),
+            "state" => $statusLookup[$status],
         ]);
 
+        $this->dataHandler->setTTL($uuid);
```
gregtyler, line 791: This is very similar to most of `CaseOutcomeCalculator`, effectively causing some duplication. I wonder if we could split that into two services like `CaseOutcomeCalculator` (which calculates the end status) and `StatusUpdateEmitter` (which sends it to Sirius).

Then this code can call the emitter and any future changes to that event integration would only need to happen in one place.

davidbridgwood: ok - so i have removed the duplication but by using the `CaseOutcomeCalculator` wherever we were previously calling the `EventSender`. Mainly because in every situation where we send the event we also need to calculate the "state" so was simpler to put it all in one service - but happy to split out if you think that is better.

service-api/module/Application/src/Model/Entity/CaseData.php
```diff
@@ -29,7 +29,7 @@ class CaseData implements JsonSerializable
     public string $id;
 
     /**
-     * @var "TELEPHONE"|"POSTOFFICE"|"VOUCHING"
+     * @var "TELEPHONE"|"POST_OFFICE"|"OnBehalf"|"cpr"
```
davidbridgwood, line 32: I've created a ticket to align enums between api/frontend so we could use the enum here rather than hard-coding values.
https://opgtransform.atlassian.net/browse/ID-542

PR #463, opened 2025-03-19
id-539 vary text on details-match page dependent on id-route/method #minor

davidbridgwood:
## Purpose

vary text on details-match page dependent on id-route/method

Fixes [id-539](https://opgtransform.atlassian.net/browse/ID-539)


davidbridgwood:
made redundant by #467

PR #465, opened 2025-03-19
id-540 vary text on vouching enter-postcode page #minor

davidbridgwood:
## Purpose

vary text on vouching enter-postcode page dependent on id-route/method

Fixes [id-540](https://opgtransform.atlassian.net/browse/ID-540)


davidbridgwood:
made redundant by #467

PR #474, opened 2025-04-01
Feature/id 546 po flow intl dln#minor

spinhalf-com:
## Purpose

Fixes ###-####

## Approach

## Learning

## Checklist


davidbridgwood (approved):
Approving but do wonder if its worth adding a quick test-case to `DocumentTypeRepositoryTest.php` to confirm the behaviour which always puts passport at the top of the list.

PR #516, opened 2025-07-24
Id 584 hmpo-api service

davidbridgwood:
## Purpose

create a service to call the hmpo-apis (auth and graphql endpoints) and wire into the application.
built against the hmpo-mock service

Fixes ###-####

## Approach

generalised `AuthApiService.php` as there is loads of common functionality with the dwp equivalent.


service-api/module/Application/test/ApplicationTest/Unit/Services/HMPO/HmpoApi/DTO/ValidatePassportRequestDTOTest.php
```diff
@@ -0,0 +1,70 @@
...
+use PHPUnit\Framework\TestCase;
+
+class ValidatePassportRequestDTOTest extends TestCase
+{
+    private array $baseCase;
+    private int $passportNumber;
+
+    public function setUp(): void
+    {
+
```
mabely, line 20: Extra line space here - just as a small finicky point

